Issue #2, opened 2018-08-05
Recursive printing of REPR List, Dictionary, Tuple


corona10:
Expected:
```
Gpython 3.4.0
>>> a = [1,2,3]
>>> a[0] = a
>>> a
[[...], 2, 3]
```

Actual behavior:
```python
Gpython 3.4.0
>>> a = [1,2,3]
>>> a[0] = a
>>> a
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

ncw:
Yes that needs fixing :-)

Want to have a go at fixing it? It probably needs repr to keep a map of visited items

corona10:
That's right approach. AFAIK, some more C API should be implemented which are related to repr_enter and repr_left :)

corona10:
@ncw Anyway, I will take a look this issue. :)

PR #3, opened 2018-08-06
py: Fix list repr on recursive element situation.

corona10:
I change the list repr not to use tuple repr, since current usage can't check an address of the parent element.

Fixes: #2

### DIFF
--- a/py/list.go
+++ b/py/list.go
@@ -2,6 +2,10 @@
 
 package py
 
+import (
+	"bytes"
+)
+
 var ListType = ObjectType.NewType("list", "list() -> new empty list\nlist(iterable) -> new list initialized from iterable's items", ListNew, nil)
 
 // FIXME lists are mutable so this should probably be struct { Tuple } then can use the sub methods on Tuple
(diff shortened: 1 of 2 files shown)


corona10:
Since this fix is not a thread-safe solution, we should decide on how to make this one will be a thread-safe.

IMHO, we have lots of choices, since we don't use reference count and GIL.

ncw:
> Since this fix is not a thread-safe solution, we should decide on how to make this one will be a thread-safe.

I haven't really thought much about thread safety so I think there are lots of things which aren't thread safe in gpython.  I've been assuming that it will be up to the user to synchronize access to mutable state...

corona10:
@ncw 
Good catch, Thanks.
I agree with thread-local storage is not a proper strategy for this project.

> Ideally you'd pass the map into each repr call

This is a best way from now on or I will find out other way :)

corona10:
@ncw I decide to implement threadState for gpython, I will send PR when the patch is ready :)

corona10:
I've tried several methods today, and your suggested method is the simplest in the current architecture, the problem is that you have to update all REPR interfaces, which is a very repetitive task.

The most elegant way seems to be to create an API to get the thread state of the current gpython interpreter. This is the way most Python interpreters do it, and the ReprEnter and ReprLeave APIs seem to be naturally implemented.

I will try as soon as possible.
Any ideas?

ncw:
Sorry for the delay in responding - I've been away with limited bandwidth!

Getting the thread state of the interpreter is going to require an ugly hack I think...  Check out
    * https://github.com/jtolds/gls **or**
    * https://github.com/tylerb/gls

The right way of doing this would be to add a `ctx context.Context` parameter to every single call which would be rather annoying!

ncw:
...or we could just punt it for the moment and use a global with a FIXME for the sake of getting things going.

PR #4, opened 2018-08-22
ci: use latest patch version of Go releases

sbinet:


ncw:
Looks great - thanks!

PR #6, opened 2018-08-22
ci: add code coverage

sbinet:

PR #8, opened 2018-08-22
gpython: point to new home of grumpy

sbinet:


sbinet:
needs #6

ncw:
I've merged this manually since github has got its knickers in a twist about conflicts!

PR #13, opened 2018-08-29
Initial work at implementing file methods

raff:
Initial work at implementing file methods:
- open (builtin) - for now only for read
- File.read
- File.write
- File.close

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -51,6 +51,7 @@ func init() {
 		// py.MustNewMethod("max", builtin_max, 0, max_doc),
 		// py.MustNewMethod("min", builtin_min, 0, min_doc),
 		py.MustNewMethod("next", builtin_next, 0, next_doc),
+		py.MustNewMethod("open", builtin_open, 0, open_doc),
 		// py.MustNewMethod("oct", builtin_oct, 0, oct_doc),
 		py.MustNewMethod("ord", builtin_ord, 0, ord_doc),
 		py.MustNewMethod("pow", builtin_pow, 0, pow_doc),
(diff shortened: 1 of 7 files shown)


ncw:
That looks very good :-)

Can you write some unit tests?  These are written in [python normally](https://github.com/go-python/gpython/tree/master/builtin/tests).  I'd suggest you make a new file `file.py`.

Not sure exactly what to do about test files - that might need a bit of thought.

raff:
Sure, will do!

raff:
retest

raff:
Do you know why the build is failing now ? In one place is complaining about gcc missing, and in the other it's failing compiling with Go 1.8 (but doesn't look it's in my code)

corona10:
@raff 

Yes, it's appveyor.yml issue.

corona10:
@raff 

You can pass the appveyor after PR #15 is merged.

corona10:
@raff 
Can you rebase your PR?

builtin/builtin.go
```diff
@@ -443,6 +444,19 @@ fromlist is not empty.  Level is used to determine whether to perform
 absolute or relative imports. 0 is absolute while a positive number
 is the number of parent directories to search relative to the current module.`
 
+const open_doc = `open(name[, mode[, buffering]]) -> file object
+
+Open a file using the file() type, returns a file object.  This is the
+preferred way to open a file.  See file.__doc__ for further information.`
+
+func builtin_open(self, obj py.Object) (py.Object, error) {
+	if filename, ok := obj.(py.String); ok {
```
sbinet: please always strive to limit indentation.
_e.g._:
```go
if filename, ok := obj.(py.String); ok {
    return py.OpenFile(string(filename))
}
return nil, py.ExceptionNewf(py.TypeError, "coercing to Unicode: need string or buffer, %v found", obj.Type().Name)
```

py/file.go
```diff
@@ -10,17 +10,126 @@
...
+	"io/ioutil"
 	"os"
 )
 
-var FileType = NewTypeX("file", `represents an open file`,
-	nil, nil)
-
 type File os.File
 
+var FileType = NewType("file", `represents an open file`)
```
sbinet: please document this exported variable.

actually.. do we need it to be exported, at this time?

py/file.go
```diff
@@ -10,17 +10,126 @@
...
+	default:
+		return nil, ExceptionNewf(TypeError, "expected a string or other character buffer object")
+	}
+
+	n, err := (*os.File)(o).Write(b)
+	return Int(n), err
+}
+
+func (o *File) readResult(b []byte) (Object, error) {
+	var asBytes = false // default mode is "as string" - also move this in File struct
```
sbinet: perhaps we could already do that as part of this CL ?

py/file.go
```diff
@@ -10,17 +10,126 @@
...
+
+	if b == nil {
+		if asBytes {
```
sbinet: please prefer early returns to limit the needed amount of indentation.

py/file.go
```diff
@@ -10,17 +10,126 @@
...
+			return Bytes{}, nil
+		} else {
+			return String(""), nil
+		}
+	}
+
+	if asBytes {
+		return Bytes(b), nil
+	} else {
```
sbinet: please prefer early returns to limit the needed amount of indentation.

py/file.go
raff: Sorry, I copied what one of the other module did and tried to follow the same pattern. Happy to update to a more gopheric style if one is defined.

But I agree that in the current implementation there are a lot of objects that are exported and maybe don't really need to be.

py/file.go
raff: Ok, I have this part done (but I am still not done parsing the options for open() so it would still just work for string output).

Found some problems in the option parser that I am trying to decide if I want to ignore them for now or trying to get them fixed.

py/file.go
ncw: @sbinet  At some point we should be running golint which warns about this.  However there is a lot of stuff to fix if we do decide to implement it!

raff:
Finished up (for now) open file and rebased.

corona10:
Please squash into the single commit please

pytest/pytest.go
```diff
@@ -12,8 +12,10 @@ import (
 	"testing"
 
 	_ "github.com/go-python/gpython/builtin"
+	_ "github.com/go-python/gpython/sys"
```
ncw: This appears to duplicate the "sys" import?

raff: Done!

raff:
After rebasing a million times (and maybe now I have finally figured out how to do it right) I think I am done!

builtin/builtin.go
```diff
@@ -437,6 +438,68 @@ fromlist is not empty.  Level is used to determine whether to perform
...
+	if encoding != py.None && encoding.(py.String) != py.String("utf-8") {
+		return nil, py.ExceptionNewf(py.NotImplementedError, "encoding not implemented yet")
+	}
+
+	if errors != py.None {
+		return nil, py.ExceptionNewf(py.NotImplementedError, "errors not implemented yet")
+	}
+
+	if newline != py.None {
+		return nil, py.ExceptionNewf(py.NotImplementedError, "errors not implemented yet")
```
corona10: Should it be like this?
```go
return nil, py.ExceptionNewf(py.NotImplementedError, "newline not implemented yet")
```

raff: good catch!

corona10:
LGTM,
IMHO, The result of code coverage issue is not important at this time.

ncw:
Notes for future work...

It would be nice to improve [the coverage of file.go](https://codecov.io/gh/go-python/gpython/compare/09f14d08f24b32e5530ee6dfa6778021c1dbcdc1...4213ccfe1fc0da0c707b920d3eb7442a52660066/src/py/file.go). Ideally what we need are tests which exercise the expected functionality of `file` and increase the coverage. The coverage shows what needs to be done!

The [args.go](https://codecov.io/gh/go-python/gpython/compare/09f14d08f24b32e5530ee6dfa6778021c1dbcdc1...4213ccfe1fc0da0c707b920d3eb7442a52660066/src/py/args.go) file doesn't have any tests which is a bit sad :-(

Issue #14, opened 2018-08-30
all([]) should return false


raff:
From python documentation at https://docs.python.org/3/library/functions.html#all
   Return True if all elements of the iterable are true (or if the iterable is empty).

corona10:
@raff 

Hi, raff on my gpython it works well.

```
➜  ~ gpython
>>> a = []
>>> all(a)
False
>>> all([])
False
>>>
```

Is there any other issue which you intended?

raff:
Sorry, I meant it should return True, according to python specs. See the link and the python implementations in there.

PR #15, opened 2018-08-30
builtin: Update builtin_all and builtin_any for Python3

corona10:
Update builtin_all and builtin_any for Python3
reference:
- https://docs.python.org/3/library/functions.html#all
- https://docs.python.org/3/library/functions.html#any

Fixes: https://github.com/go-python/gpython/issues/14

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -237,7 +237,6 @@ If the iterable is empty, return True.
 
 func builtin_all(self, seq py.Object) (py.Object, error) {
 	iter, err := py.Iter(seq)
-	res := false
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


corona10:
@sbinet 

I've also fixed appveyor.yml for solving gcc issue which was happened on #13 
PTAL

sbinet:
perhaps create a PR with only the appveyor stuff?
(so it can be merged very quickly)

corona10:
@sbinet
Okay I will create a new PR by today

sbinet:
it seems travis is in one of its bad days (it doesn't manage to find your commit...)

could you rebase on top of the latest `master`?
also, there's a typo in the CL:
`s/refernce/reference/`

corona10:
@sbinet PTAL

ncw:
I think that looks great now :-)

PR #18, opened 2018-08-30
gpython: use peterh/liner@v1.1.0

sbinet:
This CL bumps `go.mod` to use peterh/liner@v1.1.0.

Also really test Go modules by enabling `GO111MODULE=on`, only for `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
    - go: master
      env:
        - TAGS="-tags travis"
+       - GO111MODULE=on
 
 script:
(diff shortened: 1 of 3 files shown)


corona10:
@sbinet Can you squash all your commits into the single commit?

corona10:
@sbinet 
Is there any reason not to merge into one commit likewise

```
gpython: use peterh/liner@v1.1.0 and enable GO111MODULE for master
```

or is this something like a convetion for convential Golang project?
IMHO, It's too short diff for seperating into 2 commits

Issue #12, opened 2018-08-23
builtin: implement dir


sbinet:
right now, `gpython` fails with:

```python
>>> import math
>>> dir(math)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    FIXME line of source goes here
NameError: NameError: "name 'dir' is not defined"
```

PR #20, opened 2018-08-30
builtin: add dir builtin

sbinet:
Fixes go-python/gpython#12.

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -31,7 +31,7 @@ func init() {
 		py.MustNewMethod("chr", builtin_chr, 0, chr_doc),
 		py.MustNewMethod("compile", builtin_compile, 0, compile_doc),
 		// py.MustNewMethod("delattr", builtin_delattr, 0, delattr_doc),
-		// py.MustNewMethod("dir", builtin_dir, 0, dir_doc),
+		py.MustNewMethod("dir", builtin_dir, 0, dir_doc),
 		py.MustNewMethod("divmod", builtin_divmod, 0, divmod_doc),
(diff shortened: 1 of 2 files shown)


sbinet:
this is a WIP.

@ncw, for `dir()`, I need to be able to get at the content of the local scope.
how would I do that from within `builtin_dir` ?

ncw:
The way I've done that in the past, eg for implementing locals() is to implement them as InternalMethods.

Eg here is where the magic is...

https://github.com/go-python/gpython/blob/eaa7d286460777b536b404e09a1bd352f2012a78/vm/eval.go#L1578-L1600

This is where the other magic is!

```
$ git grep InternalMethodLocals
builtin/builtin.go:             py.MustNewMethod("locals", py.InternalMethodLocals, 0, locals_doc),
py/method.go:   InternalMethodLocals
vm/eval.go:             case py.InternalMethodLocals:
```

That is the only way I could find to avoid using global variables etc.  It might be too limiting eventually.

drew-512:
@sbinet after the `py.Context` merge, I could look at this out if we think `dir()` is worth it.

drew-512:
@sbinet lmk if this is worth adding/improving so we can work towards a cleared PR section.  Also, any suggestions on the next version/tag?   I think we're worthy of a 0.1 or 1.0 -- open to discussion.

sbinet:
in order to get the nice discoverability look and feel of CPython into gpython, having a working `dir()` builtin is definitely a plus.

but as we survived w/o it for almost 4 years, I don't think cleaning up this PR should be a blocker for the next tagged version of gpython.

PR #22, opened 2018-09-01
parser: Update from go tool yacc into goyacc

corona10:
Since Go 1.8 removed the go tool yacc command,
we should update it to use goyacc

reference: https://beta.golang.org/doc/go1.8#tool_yacc


corona10:
@ncw @sbinet
I notice that the code coverage metric is not exact. 
I checked that the block which said "no-cover" is already covered on the unit test.

sbinet:
I am on a train + mobile, so I can't easily check... But is the piece of code you say should be tested, is tested within the same package? (There's probably a way to treat a whole repo like a single code-coverage unit so that parts of pkgA that are tested from pkgB still are counted as tested. But I haven't found it.)

corona10:
@sbinet 
Yes, they are on the same package "parser".

ncw:
I think that looks fine to me.  I had a quick look through the auto generated code and it looks like it has just been tidied up.  That may explain the code coverage oddities also since the line count of the generated code has increased.

Issue #24, opened 2018-09-06
"print" should use __str__ or __repr__ when available


raff:
The builtin print() method is currently printing values like this (see https://github.com/go-python/gpython/blob/master/builtin/builtin.go#L192):

    fmt.Printf("%v", v)

The result is that some python objects are printed differently than expect:

None: print {}
True: print true
False: print false

The print method should probably use __str__ or __repr__ when available. As an alternative all types implementing __str__ should also implement String()

ncw:
This was fixed in #26 - thank you @raff

PR #26, opened 2018-09-07
print should use __str__ or __repr__ when available

raff:
see https://github.com/go-python/gpython/issues/24

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -189,6 +189,11 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 	end := endObj.(py.String)
 	// FIXME ignoring file and flush
 	for i, v := range args {
+		v, err := py.Str(v)
+		if err != nil {
+			return nil, err
+		}
+
 		fmt.Printf("%v", v)
 		if i != len(args)-1 {
 			fmt.Print(sep)


builtin/builtin.go
```diff
@@ -189,6 +189,14 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 	end := endObj.(py.String)
 	// FIXME ignoring file and flush
 	for i, v := range args {
+                switch sv := v.(type) {
+                    case py.I__str__:
+                        v, _ = sv.M__str__()
```
ncw: This isn't quite right - this needs to call `py.Str` which will fall back to py.Repr

If you only look for the `M__str__` then you'll miss the user defined methods.

raff: This works with this test:

    class X(object):
      def __init__(self, x):
        self.x = x
      def __str__(self):
        return "my name is %s" % self.x

    class Y(object):
      def __init__(self, x):
        self.x = x

      def __repr__(self):
        return "my nome is %s" % self.x

    print(X(42))
    print(Y('hello'))

Note that I cannot add any test to builtin/tests/builtin.py because there is no way to check the result of print (until support for redirects is implemented)

raff: re-implemented with py.Str

ncw:
I see what you mean about the tests.  I'll merge this as it is a definite improvement thanks :-)

Some tests would be nice at some point!

corona10:
@ncw In the future, we can add the test which computing sys.stdout. But not this time :)

Issue #30, opened 2018-10-09
py.Complex is missing __str__ and __repr__


raff:
Printing a complex number displays the py.Complex type instead of the real+imag values:

    >>> print(3+1j)
    <complex64 instance at %!p(py.Complex=(3+1i))>

PR #31, opened 2018-10-09
Added __str__ and __repr__ plus missing properties 

raff:
Implement https://github.com/go-python/gpython/issues/30 and add missing properties and methods
(real, imag, conjgate).

Also added tests.


ncw:
That looks great - thank you :-)

Issue #23, opened 2018-09-02
Duplicated error type.


corona10:
## Got:
``` python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    FIXME line of source goes here
TypeError: TypeError: 'all() takes exactly 1 argument (3 given)'
```

## Expected:
``` python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    FIXME line of source goes here
TypeError: 'all() takes exactly 1 argument (3 given)'
```

corona10:
@ncw 
https://github.com/go-python/gpython/blob/50cd487693e5d8e65095a2591929959c48d22d1d/vm/eval.go#L105

https://github.com/go-python/gpython/blob/50cd487693e5d8e65095a2591929959c48d22d1d/vm/eval.go#L104
To fix this issue, `vm.curexc.Type`should not be refernced since `vm.curexc.Value`
already build the exception message.

But I don't know where it works, can you guide it please?
Thanks!

ncw:
I think you want to change this in `py/exception.go`

```go
// Dump a traceback for exc to w
func (exc *ExceptionInfo) TracebackDump(w io.Writer) {
	if exc == nil {
		fmt.Fprintf(w, "Traceback <nil>\n")
		return
	}
	fmt.Fprintf(w, "Traceback (most recent call last):\n")
	exc.Traceback.TracebackDump(w)
	name := "<nil>"
	if exc.Type != nil {
		name = exc.Type.Name
	}
	fmt.Fprintf(w, "%v: %v\n", name, exc.Value)
}
```

PR #36, opened 2018-11-23
py: Fix TracebackDump not to dump duplicated exception type

corona10:
Fixes: #23


corona10:
@ncw 
PTAL, proper unit test can't add this time.
Maybe we can add capturing `fmt.Fprintf` but not the easy job to this.

ncw:
> PTAL, proper unit test can't add this time.
> Maybe we can add capturing `fmt.Fprintf` but not the easy job to this.

Well I think it is a definite improvement, so I'll merge it - thank you :-)

Issue #32, opened 2018-10-10
Errors are suppressed in generator comprehensions


ncw:
```
>>> list(i for x in range(10))
[]
```

Which should have output

```
>>> list(i for x in range(10))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <genexpr>
NameError: name 'i' is not defined
```

corona10:
@ncw I will take a look this issue

ncw:
@corona10 are you getting anywhere with this? Do you need help?

corona10:
@ncw 
https://github.com/go-python/gpython/blob/2609237ddef238242467618344ca5fb6775302e8/py/generator.go#L100
Well, I tracked the code and I found that something issue in this side. 
But I don't know which is the main reason for this issue yet.

So if you already solve this issue, Then please upload a PR for this issue :)
I can learn something from your PR.

ncw:
> Well, I tracked the code and I found that something issue in this side.
> But I don't know which is the main reason for this issue yet.

I haven't figured it out either!

But yes, it is probably something to do with making a new Vm Frame and the errors being suppressed there rather than propagated.

corona10:
@ncw 
I found the main reason for this bug.
I will upload the PR soon

ncw:
> I found the main reason for this bug.
> I will upload the PR soon

Great.  Don't forget a failing test case :-)

PR #37, opened 2018-11-24
parser: Update make_grammer_text.py

corona10:
While I work on issue #32. I found that I did not notice that 
`grammar_data_test.go` is the auto-generated file.

So I add the test cases for #22 and also implement auto-generated license file

- Support auto-generated license header based on created time.
- Add missed case when adding test cases for https://github.com/go-python/gpython/pull/22


corona10:
@ncw @sbinet PTAL

ncw:
Well spotted :-)

PR #38, opened 2018-11-28
py: Fix errors are suppressed in generator comprehensions

corona10:
Sequence.Iterate does not handle exceptions.
Iterate() should return error except StopIteration.

Fixes: https://github.com/go-python/gpython/issues/32

### DIFF
--- a/py/sequence.go
+++ b/py/sequence.go
@@ -93,9 +93,12 @@ func Iterate(obj Object, fn func(Object) bool) error {
 			return err
 		}
 		for {
-			item, finished := Next(iterator)
-			if finished != nil {
+			item, err := Next(iterator)
+			if err == StopIteration {
 				break
+			}
+			if err != nil {
+				return err
 			}
 			if fn(item) {
 				break
--- a/vm/tests/generators.py
+++ b/vm/tests/generators.py
@@ -25,6 +25,14 @@ def g2():
         yield i
 assert tuple(g2()) == (0,1,2,3,4)
 
+doc="generator 3"
+ok = False
+try:
+    list(ax for x in range(10))
+except NameError:
+    ok = True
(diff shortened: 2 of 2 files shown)


corona10:
@ncw 
I'd like to add a unit test for this.
``` python
list(ax for x in range(10))
```

Can you recommend the proper place?

corona10:
@ncw PTAL

py/sequence.go
```diff
@@ -93,10 +93,13 @@ func Iterate(obj Object, fn func(Object) bool) error {
 			return err
 		}
 		for {
-			item, finished := Next(iterator)
-			if finished != nil {
+			item, err := Next(iterator)
```
ncw: These if statements could be tidied up a bit

```go
if err == StopIteration {
    break
}
if err != nil {
    return err
}
```

vm/tests/generators.py
```diff
@@ -25,6 +25,13 @@ def g2():
         yield i
 assert tuple(g2()) == (0,1,2,3,4)
 
+doc="generator 3"
+try:
```
ncw: This needs `ok = False` before the `try:`

vm/tests/generators.py
```diff
@@ -25,6 +25,13 @@ def g2():
...
+    list(ax for x in range(10))
+except NameError:
+    ok = True,
```
ncw: There is a stray `,` on the end here!

corona10:
@ncw 
Thanks for the review. I've updated the PR.
PTAL

ncw:
That looks great thanks :-)

I'll merge it now.

It might be worth reveiwing other calls to `py.Next` as that fix looks like it was left over from a refactor when it used to return a boolean rather than an error.

Issue #41, opened 2018-12-24
runtime: Wrong implementation with range


corona10:
During I add a test for #40.
I found that this syntax does not work as I expected.

```python
a = range(100, 0, -1)
len(a) <- 100
b = [a for e in a]
len(b) <- should be 100 but 0 in gpython
```

PR #42, opened 2018-12-25
py: Fix range to support negative step

corona10:
Fixes: #41


corona10:
@ncw Please take a look :)

corona10:
@ncw 
Done! 
Thanks for the quick review, I hope you had happy Christmas with your family!

ncw:
That looks great - thanks!

Unfortunately Santa gave me and my whole family [Norovirus](https://www.nhs.uk/conditions/norovirus/) so I guess we must have been on the naughty list after all!  We've recovered now though.  I hope your Christmas was better :-)

PR #43, opened 2018-12-25
builtin: Implement enumerate feature

corona10:
Now, gpython supports enumerate feature


corona10:
@ncw PTAL

corona10:
@ncw 
It's the funniest part of this work!

ncw:
> @ncw
> It's the funniest part of this work!

Indeed, learning lots of arcane things about python!

PR #48, opened 2019-01-06
builtin: Implement min/max builtin function

corona10:
- Implement builtin min/max function
- Support the keyword-only option for `ParseTupleAndKeywords`

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -47,8 +47,8 @@ func init() {
 		// py.MustNewMethod("iter", builtin_iter, 0, iter_doc),
 		py.MustNewMethod("len", builtin_len, 0, len_doc),
 		py.MustNewMethod("locals", py.InternalMethodLocals, 0, locals_doc),
-		// py.MustNewMethod("max", builtin_max, 0, max_doc),
-		// py.MustNewMethod("min", builtin_min, 0, min_doc),
+		py.MustNewMethod("max", builtin_max, 0, max_doc),
+		py.MustNewMethod("min", builtin_min, 0, min_doc),
 		py.MustNewMethod("next", builtin_next, 0, next_doc),
 		py.MustNewMethod("open", builtin_open, 0, open_doc),
 		// py.MustNewMethod("oct", builtin_oct, 0, oct_doc),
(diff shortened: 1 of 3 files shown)


corona10:
@ncw PTAL

ncw:
Sorry, I got half way through reviewing this and got distracted!

builtin/builtin.go
```diff
@@ -772,6 +772,93 @@ func builtin_len(self, v py.Object) (py.Object, error) {
...
+
+	if positional > 1 {
+		values = args
+	} else {
+		err := py.UnpackTuple(args, nil, name, 1, 1, &values)
+		if err != nil {
+			return nil, err
+		}
+	}
+	err := py.ParseTupleAndKeywords(nil, kwargs, format, kwlist, &keyFunc, &defaultValue)
```
ncw: I note that you haven't implemented the `default` or `key` arguments...

The default argument would be very easy to implement, just copy the `defaultValue` to `maxval` if it is not None.

The `key` argument is a little trickier...

builtin/builtin.go
```diff
@@ -772,6 +772,93 @@ func builtin_len(self, v py.Object) (py.Object, error) {
...
+		item, err := py.Next(iter)
+		if err != nil {
+			if py.IsException(py.StopIteration, err) {
+				break
+			}
+			return nil, err
+		}
+		if maxval == nil {
+			//FIXME(@corona10): Should use keyfunc value
+			maxval = item
```
ncw: Duplicated line here!

builtin/builtin.go
```diff
@@ -772,6 +772,93 @@ func builtin_len(self, v py.Object) (py.Object, error) {
...
+	for {
+		item, err := py.Next(iter)
+		if err != nil {
+			if py.IsException(py.StopIteration, err) {
+				break
+			}
+			return nil, err
+		}
+		if maxval == nil {
+			//FIXME(@corona10): Should use keyfunc value
```
ncw: If key is set then doing the equivalent of `maxval = key(item)` is what is required here which shouldn't be too tricky.

corona10:
@ncw 
Looks like we have to implement keyword only arugment for `ParseTupleAndKeywords`by supporting '$'.
I will update this PR soon

corona10:
@ncw gentle ping~ :)

Issue #53, opened 2019-02-12
gpython: Implement iter()


corona10:
TODO
- [ ] Implement callable_iterator
- [ ] Implement iter() builtin

callable_itereator should be implemented before implementing iter() builtin

corona10:
````
Return a new iterator. The first parameter, callable, can be any Python callable object that can be called with no parameters; each call to it should return the next item in the iteration. When callable returns a value equal to sentinel, the iteration will be terminated.
````

PR #54, opened 2019-02-13
builtin: Implement builtin_iter

corona10:
- Implement CallIterator
- Implement builtin iter()

Closes: #53

### DIFF
--- a/py/call_iterator.go
+++ b/py/call_iterator.go
@@ -1,0 +1,51 @@
+// Copyright 2019 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// CallIterator objects
+
+package py
+
+// A python CallIterator object
+type CallIterator struct {
+	callable Object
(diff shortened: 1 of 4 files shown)


corona10:
This is WIP PR do not merge!

corona10:
@ncw 
Hi, nick please take a look if you have time!

py/call_iterator.go
```diff
@@ -0,0 +1,55 @@
...
+	return CallIteratorType
+}
+
+func (cit *CallIterator) M__iter__() (Object, error) {
+	return cit, nil
+}
+
+// Get next one from the iteration
+func (cit *CallIterator) M__next__() (Object, error) {
+	f, ok := cit.callable.(*Function)
```
ncw: You don't need to check this (and it might be a bad idea since, for instance bound methods can be called).  `py.Call` will check and return the correct TypeError

corona10:
@ncw 
Wow, thanks for the fast review!!
I 've updated the PR as you commented!

PR #55, opened 2019-02-14
builtin: Implement delattr

corona10:
I've implemented delattr for builtin function.


corona10:
@ncw 
Please take a look.
This is a pretty easy PR then before. :)

I'd like to see that gpython could run stdlib in this year.
I expect that we must update some object structure.
And this will be a rough road.

Issue #58, opened 2019-05-17
Checking dict keys fails


kellrott:
In standard python
```python
>>> x={"hello" : "world"}
>>> "hello" in x
True
```
In gpython
```python
>>> x={"hello" : "world"}
>>> "hello" in x
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    FIXME line of source goes here
TypeError: "unsupported operand type(s) for iter: 'dict'"
```

ncw:
Implementing `__iter__` for dict is very useful.

However for `in` to be efficient we should be implementing `__contains__` 

https://github.com/go-python/gpython/blob/61059b469f5b520f350d47748e8e9e9b6ccd3306/py/sequence.go#L121-L147

drew-512:
suggest @sbinet close -- addressed in https://github.com/go-python/gpython/pull/65

sbinet:
fixed w/ #65 .

PR #59, opened 2019-05-17
Adding iterator method to Dict

kellrott:
Addresses #58

### DIFF
--- a/py/dict.go
+++ b/py/dict.go
@@ -100,15 +100,13 @@ func (a StringDict) M__repr__() (Object, error) {
 	return String(out.String()), nil
 }
 
-
 func (d StringDict) M__iter__() (Object, error) {
-  o := make([]Object, 0, len(d))
-  for k := range d {
-    o = append(o, String(k))
-  }
+	o := make([]Object, 0, len(d))
+	for k := range d {
+		o = append(o, String(k))
+	}
 	return NewIterator(o), nil
 }
-
 
 func (d StringDict) M__getitem__(key Object) (Object, error) {
 	str, ok := key.(String)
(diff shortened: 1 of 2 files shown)


py/dict.go
```diff
@@ -100,6 +100,16 @@ func (a StringDict) M__repr__() (Object, error) {
 	return String(out.String()), nil
 }
 
+
+func (d StringDict) M__iter__() (Object, error) {
```
sbinet: please run `gofmt -w` on this file.

ncw:
According to the coverage report your test doesn't hit your added code?

https://codecov.io/gh/go-python/gpython/pull/59/diff

kellrott:
The code is covered in `py/tests/dict.py` line 17, but the system doesn't seem to recognize that.

py/tests/dict.py
```diff
@@ -12,4 +12,8 @@
 a = repr({"a":"b","c":5.5})
 assert a == "{'a': 'b', 'c': 5.5}" or a == "{'c': 5.5, 'a': 'b'}"
 
+doc="check __iter__"
+a = {"a":"b","c":5.5}
+assert "a" in a
```
ncw: Strictly speaking this test will no longer work when we implement the `__contains__` method for dict.

Do can you change it to something like

```python
l =  list(iter(a))
assert "a" in l
assert "c" in l
assert len(l) == 2
```

kellrott: This should be fixed now

ncw:
That looks great now thank you!  I see you added an items method too :-)

Will merge now.

PR #60, opened 2019-05-17
Adding split method to string class

kellrott:
Added `split` method to str:

```
>>> "hello|world".split("|")
['hello', 'world']
```

### DIFF
--- a/py/string.go
+++ b/py/string.go
@@ -34,6 +34,22 @@ or repr(object).
 encoding defaults to sys.getdefaultencoding().
 errors defaults to 'strict'.`, StrNew, nil)
 
+
+func init() {
+	StringType.Dict["split"] = MustNewMethod("split", func(self Object, value Object) (Object, error) {
+		selfStr := self.(String)
+		if valStr, ok := value.(String); ok {
+			ss := strings.Split(string(selfStr), string(valStr))
+			o := List{}
+			for _, j := range ss {
+				o.Items = append(o.Items, String(j))
+			}
+			return &o, nil
+		}
+		return nil, fmt.Errorf("Not split by string")
+	}, 0, "split(sub) -> split string with sub.")
+}
+
 // Type of this object
 func (s String) Type() *Type {
 	return StringType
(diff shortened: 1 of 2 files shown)


py/string.go
```diff
@@ -34,6 +34,22 @@ or repr(object).
...
+		if valStr, ok := value.(String); ok {
+			ss := strings.Split(string(selfStr), string(valStr))
+			o := List{}
+			for _, j := range ss {
+				o.Items = append(o.Items, String(j))
+			}
+			return &o, nil
+		}
+		return nil, fmt.Errorf("Not split by string")
+	}, 0, "split(sub) -> split string with sub.")
```
ncw: The Python help is

```
    S.split([sep [,maxsplit]]) -> list of strings
    
    Return a list of the words in the string S, using sep as the
    delimiter string.  If maxsplit is given, at most maxsplit
    splits are done. If sep is not specified or is None, any
    whitespace string is a separator and empty strings are removed
    from the result.
```

Which makes me see that we are missing two things.
- If `sep` isn't passed (or is None) in then we should be using `strings.Fields` in go terms (This doesn't have a maxfields parameter though.)
- There is another optional parameter to specify the number of splits.

Here are some cases to consider

```
>>> "a,d,c".split(",")
['a', 'd', 'c']
>>> "a,d,c".split(",",1)
['a', 'd,c']
>>> " a   d   b   ".split()
['a', 'd', 'b']
>>> " a   d   b   ".split(None, 1)
['a', 'd   b   ']
```

py/tests/string.py
```diff
@@ -99,6 +99,9 @@ class C():
 asc="hello"
 uni="£100世界𠜎" # 1,2,3,4 byte unicode characters
 
+doc="split"
+assert 4 == len(list("0,1,2,4".split(",")))
```
ncw: I think `assert ["0","1","2","4"] == "0,1,2,4".split(",")` would be better here.  A few more tests would be good too, like for empty string.

kellrott:
The error being thrown in the unit test is the older text from the exception, like the copy being tested is from an older build. Is it possible to clear the cache and re-try the unit test?

ncw:
That looks great now thank you and 100% diff coverage too :-)

I'll merge it now - thank you very much.

PR #61, opened 2019-06-28
String and List Methods

kellrott:
This PR adds
 - `string.startswith`
 - `string.endswith`
 - `list.append`
 - `list.extend`


ncw:
That looks great thank you :-)

Will merge now.

Issue #62, opened 2019-07-18
no such file or directory at first time gpython repl


msAlcantara:
I installed gpython with the command `go get github.com/go-python/gpython` and when I run in the first time in repl I get the next message error:
`Failed to open history: open /home/matheus/.gpyhistory: no such file or directory`
the repl do not stop working, and not presented no problem at first, but should be nice dont show this kind of error since that this does not hassles the gpython experience repl

ncw:
I think this would be quite easy to fix - fancy having a go?

msAlcantara:
Yeah, I can do this

PR #63, opened 2019-07-22
fix: ~/.gpyhistory: no such file or directory

msAlcantara:
Now when the ~/.gpyhistory file does not exist, do not show the error
message in the repl, as long as this does not disrupt the experience in the repl.

Fixes go-python/gpython#62.


ncw:
That is great - perfect fix, thank you.

PR #66, opened 2019-07-26
builtin: Implement builtin_ascii

corona10:
Implement builtin_ascii

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -24,7 +24,7 @@ func init() {
 		py.MustNewMethod("abs", builtin_abs, 0, abs_doc),
 		py.MustNewMethod("all", builtin_all, 0, all_doc),
 		py.MustNewMethod("any", builtin_any, 0, any_doc),
-		// py.MustNewMethod("ascii", builtin_ascii, 0, ascii_doc),
+		py.MustNewMethod("ascii", builtin_ascii, 0, ascii_doc),
 		// py.MustNewMethod("bin", builtin_bin, 0, bin_doc),
 		// py.MustNewMethod("callable", builtin_callable, 0, callable_doc),
 		py.MustNewMethod("chr", builtin_chr, 0, chr_doc),
@@ -307,6 +307,19 @@ func builtin_any(self, seq py.Object) (py.Object, error) {
 		}
 	}
 	return py.False, nil
+}
+
+const ascii_doc = `
+`
+
+func builtin_ascii(self, o py.Object) (py.Object, error) {
+	reprObj, err := py.Repr(o)
+	if err != nil {
+		return nil, err
+	}
+	repr := reprObj.(py.String)
+	out := py.StringEscape(repr, true)
+	return py.String(out), err
 }
 
 const round_doc = `round(number[, ndigits]) -> number
--- a/builtin/tests/builtin.py
+++ b/builtin/tests/builtin.py
@@ -19,6 +19,12 @@ assert any((1,1,0)) == True
 assert any(["hello", "world"]) == True
 assert any([]) == False
 
+doc="ascii"
+assert ascii('hello world') == "'hello world'"
+assert ascii('안녕 세상') == "'\\uc548\\ub155 \\uc138\\uc0c1'"
+assert ascii(chr(0x10001)) == "'\\U00010001'"
+assert ascii('안녕 gpython') == "'\\uc548\\ub155 gpython'"
+
 doc="chr"
 assert chr(65) == "A"
 assert chr(163) == "£"
--- a/py/string.go
+++ b/py/string.go
@@ -35,6 +35,61 @@ or repr(object).
 encoding defaults to sys.getdefaultencoding().
 errors defaults to 'strict'.`, StrNew, nil)
 
+// Escape the py.String
+func StringEscape(a String, ascii bool) string {
+	s := string(a)
+	var out bytes.Buffer
+	quote := '\''
+	if strings.ContainsRune(s, '\'') && !strings.ContainsRune(s, '"') {
+		quote = '"'
+	}
+	if !ascii {
+		out.WriteRune(quote)
+	}
+	for _, c := range s {
+		switch {
+		case c < 0x20:
+			switch c {
+			case '\t':
+				out.WriteString(`\t`)
+			case '\n':
+				out.WriteString(`\n`)
+			case '\r':
+				out.WriteString(`\r`)
+			default:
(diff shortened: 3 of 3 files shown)


corona10:
@ncw 
Please take a look!

builtin/tests/builtin.py
```diff
@@ -19,6 +19,10 @@
 assert any(["hello", "world"]) == True
 assert any([]) == False
 
+doc="ascii"
+assert ascii('hello world') == "'hello world'"
+assert ascii('안녕 세상') == "'\\uc548\\ub155 \\uc138\\uc0c1'"
```
ncw: I think this needs a few more test cases, and you'll see that there are some cases missing from the above

```
>>> ascii("\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0A\x0B\x0C\x0D\x0E\x0F\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1A\x1B\x1C\x1D\x1E\x1F")
"'\\x00\\x01\\x02\\x03\\x04\\x05\\x06\\x07\\x08\\t\\n\\x0b\\x0c\\r\\x0e\\x0f\\x10\\x11\\x12\\x13\\x14\\x15\\x16\\x17\\x18\\x19\\x1a\\x1b\\x1c\\x1d\\x1e\\x1f'"
>>> print(ascii("\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0A\x0B\x0C\x0D\x0E\x0F\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1A\x1B\x1C\x1D\x1E\x1F"))
'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f'
>>> print(ascii("'hello'"))
"'hello'"
>>> print(ascii('"hello"'))
'"hello"'
>>> print(ascii('\\'))
'\\'
```

In fact you could do worse than adapting the tests here.

https://github.com/go-python/gpython/blob/bd0985be3562d0a36eef5e34ecae4d4a4235fd57/py/tests/string.py#L35

Note that this should probably call `ascii()` on some non strings too.

builtin/builtin.go
```diff
@@ -309,6 +312,37 @@ func builtin_any(self, seq py.Object) (py.Object, error) {
...
+	reprObject, err := py.Repr(o)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(corona10): Replace it into strings.Builder
+	var sb bytes.Buffer
+	repr := reprObject.(py.String)
+	for _, c := range repr {
+		if c <= unicode.MaxASCII {
```
ncw: This needs to \xXX encode chars in the range 0..31 inclusive with exceptions for \t \r \n

builtin/builtin.go
```diff
@@ -309,6 +312,37 @@ func builtin_any(self, seq py.Object) (py.Object, error) {
...
+const ascii_doc = `ascii(obj, /)
+Return an ASCII-only representation of an object.
+
+As repr(), return a string containing a printable representation of an
+object, but escape the non-ASCII characters in the string returned by
+repr() using \\x, \\u or \\U escapes. This generates a string similar
+to that returned by repr() in Python 2
+`
+
+func builtin_ascii(self, o py.Object) (py.Object, error) {
```
ncw: I note that this has already been implemented **almost** here

https://github.com/go-python/gpython/blob/bd0985be3562d0a36eef5e34ecae4d4a4235fd57/py/string.go#L196

So you could factor that into its own function `stringEscape(s string, ascii bool)` with an extra parameter `ascii` which if set would encode everything, so wherever it says `strconv.IsPrint(c)` in that function you'd say `!ascii && strconv.IsPrint(c)`

That function deals with \r \t \n, the correct escaping for big chars and `'"'` vs `"'"`

builtin/builtin.go
corona10: @ncw Oh, that's interesting!

builtin/builtin.go
corona10: nice!

builtin/builtin.go
corona10: @ncw 

Could you provide me a sample python code to reproduce it, please?
Unfortunately, I can not reproduce it..

builtin/builtin.go
corona10: Unfortunately, implementing `stringEscape(s string, ascii bool)` is not properly work in an easy case for this API.

for example
```
ascii('hello world')
# should be "'hello world'"
# but got '"\'hello world\'"'
```

corona10:
@ncw 
Thanks for the kind review.
I 've added additional tests.
But I didn't reflect some of your reviews that I 've not understood yet.
I left comments on that!
Please take a look.

builtin/builtin.go
corona10:
```python
# cpython 3.7
>>> ascii(chr(30))
"'\\x1e'"

# current PR

>>> ascii(chr(30))
"'\\x1e'"
```

corona10:
I've found this code does not work well.
```python
>>> ascii('\U+10001')
"'\\U00010001'"
```

builtin/builtin.go
ncw: Note that if you are using the REPL then you want to print() the output otherwise it will be double escaped which looks like what has happened above.

I think using the `stringEscape` as suggested should fix the problems!  It will also fix quite a few corner cases you haven't considered yet!

corona10:
@ncw 
I've added StringEscape which can be used for __repr__ or ascii.
Note that StringEscape which need to ascii is below code.

If there is something I missed. Please let me know!

````go

func stringEscape(a py.String) string {
    s := string(a)
    var out bytes.Buffer
    for _, c := range s {
        switch {
        case c < 0x20:
            switch c {
            case '\t':
                out.WriteString(`\t`)
            case '\n':
                out.WriteString(`\n`)
            case '\r':
                out.WriteString(`\r`)
            default:
                out.WriteRune(c)
            }
        case c < 0x100:
            out.WriteRune(c)
        case c < 0x10000:
            fmt.Fprintf(&out, "\\u%04x", c)
        default:
            fmt.Fprintf(&out, "\\U%08x", c)
        }
    }
    return out.String()
}
````

ncw:
LGTM - thank you - please merge :-)

PR #67, opened 2019-07-27
py: Fix __mul__ of list and tuple on negative case

corona10:
py: Fix __mul__ of list and tuple on negative case


corona10:
@ncw PTAL

PR #70, opened 2019-08-12
builtin: Implement builtin_bin

corona10:
Implementation of builtin bin


corona10:
@ncw please take a look

PR #74, opened 2019-09-10
Generate SyntaxError of global declaration

### LINKED ISSUE #72: SyntaxError of global declaration don't stop program
```python
a = 3
global a
print(a)

b = 2 + 5
print(b)
```
This code generate syntax error, but gpython doesn't stop program

### Expected Result
```bash
Hyeockz:bin hyeockjinkim$ python3 g.py 
  File "g.py", line 2
    global a
SyntaxError: name 'a' is assigned to before global declaration
```

### Actual Result
```bash
2019/09/09 00:08:36 name 'a' is assigned to before global declaration
3
7
```
SyntaxError in python3 is Error, not warning. So I think it should be modified.

HyeockJinKim:
Generate SyntaxError instead of SyntaxWarning
if global declaration is for priviously used variable

Fixes #72


corona10:
@HyeockJinKim 
Raising SyntaxError is a right solution but
It looks like tests should be updated

corona10:
@ncw 
Would you like to guide about updating the unit test for him?
If his approach is not proper, feedback will be helpful to him

HyeockJinKim:
@corona10 Can I get a review?

ncw:
I think the code looks nice :-)

**However** it makes gpython incompatible with python3.4 as I wrote here: https://github.com/go-python/gpython/issues/72#issuecomment-529222858

Now I don't think any of the tests rely on that so it is safe to change it.

What do you think @corona10 - make a break with python3.4 compatibility here? On balance I think it is a good idea since this is almost always an actual problem.

PS the commits need squashing which can be done at merge time

corona10:
@ncw 
I am not on +1 or -1 on breaking the 3.4 compatibilities.
However when we update the project objective into implementing Python3.x interpreter.
Breaking with python3.4 compatibility is not a bad idea if so that's a way we have to go.

ncw:
> @ncw
> I am not on +1 or -1 on breaking the 3.4 compatibilities.
> However when we update the project objective into implementing Python3.x interpreter.
> Breaking with python3.4 compatibility is not a bad idea if so that's a way we have to go.

OK, lets squash and merge :-)

corona10:
@HyeockJinKim
Thanks for the contribution
and thanks for the review @ncw  :)

PR #81, opened 2019-09-15
Add `sorted` and `list.sort`

tim-st:


tim-st:
@corona10 Ok, I will add some test cases.
There is a PR at golang that implements [Block Sort](https://en.wikipedia.org/wiki/Block_sort), which will improve the stable sort, once it's merged. Timsort isn't wanted in Go because it uses O(n) space currently.

corona10:
@tim-st Thanks for the information!

corona10:
@ncw @sbinet Can you take a look at this PR?

tim-st:
I will add one more thing.
```python
a = [1,3,2]
assert list.sort(a) is None
assert a == [1, 2, 3]
```
should work. Currently this syntax doesn't work for `list.append` etc too.

tim-st:
@ncw Thanks! I'd like to change it to get a higher score, but I don't understand how "codecov" measures the results, and there are no hints on the page which parts should be changed :\

ncw:
> @ncw Thanks! I'd like to change it to get a higher score, but I don't understand how "codecov" measures the results, and there are no hints on the page which parts should be changed :\

:-)

The output is a bit cryptic

The easiest way to improve the coverage is to use the [go coverage tools](https://golang.org/cmd/cover/) locally to see the bits which haven't been covered in the code you've added.  Cover those and the coveralls score will increase!

tim-st:
Hm, when I run this `cover` tool, most of the code in `list.go` is marked red, not only my code. I don't know how to improve this, maybe someone can show me how to get this right.

corona10:
> This looks like excellent work :-)
> If we could get the coverage of the patch up to above the threshold that would be perfect!
> I didn't see anything which I thought needed changing :-

@tim-st  cc @ncw 
If there no issue with the current code without code coverage.
We can go to merge it.

corona10:
@tim-st 
Thanks for your contribution!

Issue #76, opened 2019-09-10
Implement set operation


corona10:
These operations should be implemented

``` python
>>> a = {1, 2, 3}
>>> b = {2, 3, 4, 5}
>>> a | b
{1, 2, 3, 4, 5}
>>> a & b
{2, 3}
>>> a - b
{1}
>>> a ^ b
{1, 4, 5}
>>>
```

corona10:
These operations can be implemented on here based on below code.

https://github.com/go-python/gpython/blob/master/py/set.go#L65

````go
func (s *Set) M__and__(other Object) (Object, error) {
        // TODO: Implment
       return ret, nil
}
````

DoDaek:
I want to implement this issue

corona10:
@DoDaek 
I 'd like to recommend implementing `M__and__` first.

ncw:
I wrote the set primitives for gotemplate here: https://github.com/ncw/gotemplate/blob/master/set/set.go

These were based off the python primitives so should work with very little modification other than changing the types.

Feel free to use those here :-)

SanggiHong:
I found another operation to implement.
```python
// python 3.7.4
>>> a = set([1,2,3])
>>> for ai in a:
...   print(ai)
...
1
2
3
```python
// gpython
>>> a = set([1,2,3])
panic: interface conversion: py.Object is *py.List, not py.Tuple

goroutine 1 [running]:
github.com/go-python/gpython/py.SetNew(0xc0000ed290, 0xc00013ead0, 0x1, 0x3, 0x0, 0x14bc828, 0x0, 0x0, 0x42152e0)
	/Users/sanggihong/go/src/github.com/go-python/gpython/py/set.go:63 +0x20d
github.com/go-python/gpython/py.(*Type).M__call__(0xc0000ed290, 0xc00013ead0, 0x1, 0x3, 0x0, 0xc000115301, 0x104491e, 0xc0000f0ec0, 0xc00009db70)
	/Users/sanggihong/go/src/github.com/go-python/gpython/py/type.go:380 +0x82
github.com/go-python/gpython/py.Call(0x12f9f40, 0xc0000ed290, 0xc00013ead0, 0x1, 0x3, 0x0, 0x1279040, 0x1, 0xc0002a1e40, 0xc000115448)
	/Users/sanggihong/go/src/github.com/go-python/gpython/py/internal.go:169 +0x299
[...]

PR #82, opened 2019-09-15
Implement and of set

DoDaek:
ISSUE: #76

### DIFF
--- a/py/set.go
+++ b/py/set.go
@@ -111,6 +111,20 @@ func (s *Set) M__iter__() (Object, error) {
 	return NewIterator(items), nil
 }
 
+func (s *Set) M__and__(other Object) (Object, error) {
+	ret := NewSet()
+	b, ok := other.(*Set)
+	if !ok {
+		return nil, TypeError
+	}
+	for i := range b.items {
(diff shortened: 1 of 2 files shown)


py/tests/set.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 The go-python Authors.  All rights reserved.
```
corona10: Copyright 2018 -> Copyright 2019

py/set.go
```diff
@@ -111,6 +111,20 @@ func (s *Set) M__iter__() (Object, error) {
 	return NewIterator(items), nil
 }
 
+func (s *Set) M__and__(other Object) (Object, error) {
+	ret := NewSet()
+	b, ok := other.(*Set)
+	if !ok {
+		return nil, TypeError
```
corona10:
```go
return nil, ExceptionNewf(TypeError, "unsupported operand type(s) for &: '%s' and '%s'", s.Type().Name, other.Type().Name)
```

DoDaek: I change the return value of the function

PR #83, opened 2019-09-15
Add Slice function for range type

### LINKED ISSUE #77: Slice function of range type must be supported
```python
a = range(10)
d = a[10:11]
print(d)
```
Unlike the list type, the range type does not support the slice function. slice in the range type must be supported.

### Expected result
```bash
range(10, 10)
```

### Actual result
```bash
TypeError: "unsupported operand type(s) for index: 'slice'"
2019/09/11 23:10:28 TypeError: "unsupported operand type(s) for index: 'slice'"
```

HyeockJinKim:
Crate a new range object by calculating start,
stop, and step when slice is entered as argument
to the __getitem__ function of the range

Fixes #77

### DIFF
--- a/py/range.go
+++ b/py/range.go
@@ -79,18 +79,18 @@ func RangeNew(metatype *Type, args Tuple, kwargs StringDict) (Object, error) {
 }
 
 func (r *Range) M__getitem__(key Object) (Object, error) {
+	if slice, ok := key.(*Slice); ok {
+		return computeRangeSlice(r, slice)
+	}
+
 	index, err := Index(key)
 	if err != nil {
 		return nil, err
 	}
-	// TODO(corona10): Support slice case
-	length := computeRangeLength(r.Start, r.Stop, r.Step)
-	if index < 0 {
-		index += length
-	}
+	index = computeNegativeIndex(index, r.Length)
 
-	if index < 0 || index >= length {
-		return nil, ExceptionNewf(TypeError, "range object index out of range")
+	if index < 0 || index >= r.Length {
+		return nil, ExceptionNewf(IndexError, "range object index out of range")
 	}
 	result := computeItem(r, index)
 	return result, nil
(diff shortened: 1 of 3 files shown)


py/range.go
```diff
@@ -79,12 +79,54 @@ func RangeNew(metatype *Type, args Tuple, kwargs StringDict) (Object, error) {
 }
 
 func (r *Range) M__getitem__(key Object) (Object, error) {
+	length := r.Length
+	if slice, ok := key.(*Slice); ok {
+		length, err := length.GoInt()
```
corona10: Please extract this code lines into a single private method.

tim-st:
CPython (v3.7.4) has an additional type check that raises an exception on wrong types like:
`TypeError: slice indices must be integers or None or have an __index__ method`.
That means types having `__index__` method operate on this.
For example:
```python
class C:
    def __index__(self):
        return 1
r = range(10)
assert len(r[:C()]) == 1
# len(r[:1.1]) -> Exception
```

corona10:
@HyeockJinKim Please rebase the PR and reflect @tim-st 's review.

HyeockJinKim:
@tim-st @corona10 
Can I get a review again?

tim-st:
@HyeockJinKim 
Your code for `__index__` doesn't check the return type.

For example:

```python
class C:
    def __index__(self):
       return "abc"
r[:C()]
```
Should raise `TypeError: __index__ returned non-int (type str)`.

Maybe it would be good to first call `__index__` using `py.Call` and then check error and then the type of the returned result if no error.

Also the following should throw an Exception like in CPython:
`r[:1.1] # TypeError: slice indices must be integers or None or have an __index__ method`

But I think that is quite much work to implement, so we could also just handle this as a new issue.

The following code results are different in CPython (sorry, I didn't find this before):
```Python
r = range(10)
r[-2::-2] # r[-2::2] is ok 
r[-2::-1]
r[:-2:-1]
r[-2:-2:-2]
```

PR #84, opened 2019-09-17
set: Implement __or__ of set

DoDaek:
Implement __or__ of set
ISSUE: #76


py/set.go
```diff
@@ -125,6 +125,20 @@ func (s *Set) M__and__(other Object) (Object, error) {
 	return ret, nil
 }
 
+func (s *Set) M__or__(other Object) (Object, error) {
+	ret := s
```
corona10: You should return new set instance.

DoDaek: I'll make corrections on that part.

PR #87, opened 2019-09-20
Implement range object

### LINKED ISSUE #86: Implement range object
## Expected Result
```python
>>> range(10)
range(0, 10)
>>> range(10) == range(10)
True
>>> str(range(10))
'range(0, 10)'
```

## Actual Result
```python
>>> range(10)
<range instance at 0xc0002c8380>
>>> range(10) == range(10)
TypeError: "unsupported operand type(s) for ==: 'range' and 'range'"
>>> str(range(10))
'<range instance at 0xc0002c8c60>'
```
The `__repr__`, `__str__`, `__eq__`, and `__ne__` functions of the `range` object are not implemented.

HyeockJinKim:
Implement `__repr__`, `__str__`, `__eq__`, `__ne__` of range

Fixes #86

### DIFF
--- a/py/range.go
+++ b/py/range.go
@@ -196,3 +196,58 @@ func computeRangeLength(start, stop, step Int) Int {
 var _ I__getitem__ = (*Range)(nil)
 var _ I__iter__ = (*Range)(nil)
 var _ I_iterator = (*RangeIterator)(nil)
+
+
+func (a *Range) M__eq__(other Object) (Object, error) {
+	b, ok := other.(*Range)
+	if !ok {
+		return NotImplemented, nil
+	}
+
+	if a.Length != b.Length {
+		return False, nil
+	}
+
+	if a.Length == 0 {
+		return True, nil
+	}
+	if a.Start != b.Start {
+		return False, nil
+	}
+
+	if a.Step == 1 {
+		return True, nil
+	}
+	if a.Step != b.Step {
+		return False, nil
+	}
+
+	return True, nil
+}
+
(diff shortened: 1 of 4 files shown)


py/range.go
```diff
@@ -104,6 +106,42 @@ func (r *Range) M__iter__() (Object, error) {
 	}, nil
 }
 
+func (r *Range) M__str__() (Object, error) {
+	return r.M__repr__()
+}
+
+func (r *Range) repr() (Object, error) {
+	var out bytes.Buffer
```
sbinet: consider using [strings.Builder](https://godoc.org/strings#Builder) instead? (needs Go >= 1.10, though)

corona10: strings.Builder is more recommended way :)

py/range.go
```diff
@@ -104,6 +108,42 @@ func (r *Range) M__iter__() (Object, error) {
 	}, nil
 }
 
+func (r *Range) M__str__() (Object, error) {
+	return r.M__repr__()
+}
+
+func (r *Range) repr() (Object, error) {
+	var b strings.Builder
```
corona10: Since strings.Builder is not supported at go < 1.10
You need to use version constraints.
Read this article. https://medium.com/@theckman/version-constraints-and-go-c9309be15773

example: 
https://github.com/corona10/goimagehash/blob/master/imagehash18.go
https://github.com/corona10/goimagehash/blob/master/imagehash19.go

py/range.go
HyeockJinKim: I modified the code. Would you please review it again?

PR #88, opened 2019-09-22
set: Implement __sub__ and __xor__ of set

DoDaek:
ISSUE : #76


corona10:
Thanks for the contribution!

PR #90, opened 2019-09-25
Fix "end" option in print func

Sungmin-Joo:
line 190 to 193.
I tried to apply the end option to the print function.
Added code performed when "end" value of parameter "kwargs" is not "nil".

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -176,10 +176,10 @@ flush: whether to forcibly flush the stream.`
 
 func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Object, error) {
 	var (
-		sepObj py.Object = py.String(" ")
+	  	sepObj py.Object = py.String(" ")
 		endObj py.Object = py.String("\n")
 		file   py.Object = py.MustGetModule("sys").Globals["stdout"]
-		flush  py.Object
+	 	flush  py.Object
 	)
 	kwlist := []string{"sep", "end", "file", "flush"}
 	err := py.ParseTupleAndKeywords(nil, kwargs, "|ssOO:print", kwlist, &sepObj, &endObj, &file, &flush)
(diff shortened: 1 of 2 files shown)


builtin/builtin.go
```diff
@@ -176,17 +176,21 @@ flush: whether to forcibly flush the stream.`
 
 func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Object, error) {
 	var (
-		sepObj py.Object = py.String(" ")
+	  	sepObj py.Object = py.String(" ")
```
corona10: @Sungmin-Joo What is this changed?

builtin/builtin.go
```diff
@@ -176,17 +176,21 @@ flush: whether to forcibly flush the stream.`
...
 		endObj py.Object = py.String("\n")
 		file   py.Object = py.MustGetModule("sys").Globals["stdout"]
-		flush  py.Object
+	 	flush  py.Object
```
corona10: Looks like unnecessary changed

builtin/builtin.go
```diff
@@ -195,7 +199,7 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 	}
 
 	for i, v := range args {
-		v, err := py.Str(v)
+	 	v, err := py.Str(v)
```
corona10: same here

builtin/builtin.go
```diff
@@ -209,7 +213,7 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 			_, err = py.Call(write, py.Tuple{sep}, nil)
 			if err != nil {
 				return nil, err
-			}
+	 		}
```
corona10: same here

tim-st:
@Sungmin-Joo 
The following code writes different string to the stream than in CPython:
`print(1,2,3, end="||\n")`

PR #92, opened 2019-09-27
Fixed "print function's "end" option error in CPython"

Sungmin-Joo:
The following code writes different string to the stream than in CPython:
```
print(1,2,3, end="||\n")
```

I fixed that bug by modifying the code
resolved: #93

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -186,6 +186,12 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 	if err != nil {
 		return nil, err
 	}
+
+	if kwargs["sep"] != nil {
+		sepObj = kwargs["sep"]
+	} else {
+		sepObj = py.String(" ")
+	}
 	sep := sepObj.(py.String)
 
 	if kwargs["end"] != nil {
--- a/builtin/tests/builtin.py
+++ b/builtin/tests/builtin.py
@@ -282,6 +282,12 @@ except TypeError as e:
 assert ok, "TypeError not raised"
 
 try:
+    print("hello","gpython", end="123\n")
+except TypeError as e:
+    ok = True
+assert ok, "TypeError not raised"
+
+try:
     print("hello", sep=" ", end=1)
 except TypeError as e:
     #if e.args[0] != "end must be None or a string, not int":


Sungmin-Joo:
@corona10  
 Is it a proper test?
If not, I will try again tomorrow.

corona10:
@Sungmin-Joo 
You should show the test
## AS-IS
```python
>>> print(1,2,3, end="||\n")
1||
2||
3||
```

## TO-BE

```python
>>> print(1,2,3, end="||\n")
1 2 3||
```

builtin/builtin.go
```diff
@@ -186,6 +186,12 @@ func builtin_print(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Obje
 	if err != nil {
 		return nil, err
 	}
+
+	if kwargs["sep"] != nil {
```
sbinet: what about:
```go
switch sep, ok := kwargs["sep"]; ok {
case true:
    sepObj = sep
default:
    sepObj = py.String(" ")
}
```

builtin/tests/builtin.py
```diff
@@ -281,6 +281,12 @@ def gen2():
     ok = True
 assert ok, "TypeError not raised"
 
+try:
+    print("hello","gpython", end="123\n")
+except TypeError as e:
+    ok = True
+assert ok, "TypeError not raised"
```
sbinet:
```suggestion
assert ok, "TypeError not raised when 'sep' is not in the kwargs dict"
```

that said, I am not sure I understand this test.
do we really want this line to raise an exception? why?

builtin/tests/builtin.py
```diff
@@ -281,6 +281,12 @@ def gen2():
     ok = True
 assert ok, "TypeError not raised"
 
+try:
+    print("hello","gpython", end="123\n")
```
sbinet:
```suggestion
    print("hello", "gpython", end="123\n")
```

corona10:
@Sungmin-Joo @sbinet 
Sorry, 
I think that we should fix #90 was not the proper approach.
```go
err := py.ParseTupleAndKeywords(nil, kwargs, "|ssOO:print", kwlist, &sepObj, &endObj, &file, &flush)
```
I am now closing this PR
Let's fix this issue with proper approach :)

PR #94, opened 2019-09-28
Revert "Fix "end" option in print func"

corona10:
Reverts go-python/gpython#90

Sorry guys, I didn't review the patch #90 well.
The patch was not the proper approach.


corona10:
cc @Sungmin-Joo

PR #97, opened 2019-09-28
Handle the non-integer return of __index__

### LINKED ISSUE #96: Handle the return of non-integer value in __index__ function
An error should occur when returning a non-integer value from __index__.

### Expected result (cpython)
```python
>>> class C:
...     def __index__(self):
...         return 'a'
... 
>>> c = C()
>>> a = [1, 2 ,4]
>>> a[c]
TypeError: __index__ returned non-int (type str)
```

### Actual result (gpython)
```python
>>> class C:
...     def __index__(self):
...         return 'a'
... 
>>> c = C()
>>> a = [1, 2, 4]
>>> a[c]
1
```

HyeockJinKim:
Generate TypeError when __index__ return non-integer value

Fixes #96

### DIFF
--- a/py/internal.go
+++ b/py/internal.go
@@ -95,11 +95,12 @@ func Index(a Object) (Int, error) {
 		if err != nil {
 			return 0, err
 		}
+
 		if res, ok := A.(Int); ok {
 			return res, nil
 		}
 
-		return 0, err
+		return 0, ExceptionNewf(TypeError, "__index__ returned non-int: (type %s)", A.Type().Name)
 	}
 
 	return 0, ExceptionNewf(TypeError, "unsupported operand type(s) for index: '%s'", a.Type().Name)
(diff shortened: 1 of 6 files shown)


HyeockJinKim:
@corona10 
Can I get a review?

tim-st:
Does this also affect the `list` type?
CPython supports:
```python
>>> l = [1,2]
>>> class C:
...   def __index__(self):
...     return 1
...
>>> l[C()]
2
```
If your change has an influence on this, you could add a test for list, too.

HyeockJinKim:
@tim-st 
This fix affects strings, tuples, and lists. 
So I added test code for each one. Would you like to review it again?

tim-st:
@HyeockJinKim Looks good, I think it should work now!

ncw:
That looks great now - thank you :-)

Issue #98, opened 2019-09-29
Implement slice object


HyeockJinKim:
Currently the slice object is not fully implemented.
Many features, including instance creation, are not fully implemented and must be implemented.

https://github.com/go-python/gpython/blob/8c361a8eb99395b87df36de243df7f4d7b8d07d9/builtin/builtin.go#L91
https://github.com/go-python/gpython/blob/8c361a8eb99395b87df36de243df7f4d7b8d07d9/py/slice.go#L16-L18

### Expected result (cpython)
```python 
>>> slice(10)
slice(None, 10, None)
>>> slice(10).indices(0)
(0, 0, 1)
```

PR #99, opened 2019-09-29
Add __new__ function and property of slice

HyeockJinKim:
Issue #98


HyeockJinKim:
@corona10 
Can I get a review?

builtin/testfile
```diff
@@ -0,0 +1 @@
+1,2,3,
```
corona10: What is this file?

HyeockJinKim: The file seems to be uploaded incorrectly. I deleted it.

PR #100, opened 2019-09-29
set: Implement initialization set with sequence

SanggiHong:
Implement the initialization of Set with a sequence. looked up the implementation of the List
issue #76

### DIFF
--- a/py/tests/set.py
+++ b/py/tests/set.py
@@ -60,6 +60,9 @@ doc="set"
 a = set([1,2,3])
 b = set("set")
 c = set((4,5))
+assert len(a) == 3
+assert len(b) == 3
+assert len(c) == 2
 assert 1 in a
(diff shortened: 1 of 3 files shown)


py/tests/set.py
```diff
@@ -55,6 +55,17 @@
...
 d = a ^ b
 assert 1 in c
+
+doc="set"
+a = set([1,2,3])
+b = set("set")
+c = set((4,5))
+assert 1 in a
+assert 2 in a
+assert 3 in a
```
corona10: @SanggiHong 
Would you like to add below codes for the test? 
```python
assert len(a) == 3
assert len(b) == 3
assert len(c) == 2
```

SanggiHong: No problem, just added :)

Issue #103, opened 2019-09-29
`repr([1, 1.0])` is different from CPython


tim-st:
CPython:
`repr([1,1.0]) # '[1, 1.0]'` 

GPython:
`repr([1,1.0]) # '[1, 1]'`

PR #104, opened 2019-09-29
Change repr(float) if float(int(f)) == f

tim-st:
Fixes https://github.com/go-python/gpython/issues/103


ncw:
Thank you very much :-)

PR #106, opened 2019-09-29
Implementing the "get" function on a "Dictionary"

### LINKED ISSUE #105: The "get" function is not implemented in the Dict.
in gpython
``` python
>>>> d = {'a' : 1}
>>>> d.get('a')
  File "test.py", line 13, in <module>
AttributeError: "'dict' has no attribute 'get'"
2019/09/29 16:24:04 AttributeError: "'dict' has no attribute 'get'"
```

in python
``` python
>>>> d = {'a' : 1}
>>>> d.get('a')
1
```

Sungmin-Joo:
Implement the "get" function and add unit test.
fixes #105

### DIFF
--- a/py/dict.go
+++ b/py/dict.go
@@ -36,6 +36,30 @@ func init() {
 		}
 		return NewIterator(o), nil
 	}, 0, "items() -> list of D's (key, value) pairs, as 2-tuples")
+
+	StringDictType.Dict["get"] = MustNewMethod("get", func(self Object, args Tuple) (Object, error) {
+		var length = len(args)
+		switch {
+		case length == 0:
+			return nil, ExceptionNewf(TypeError, "%s expected at least 1 arguments, got %d", "items()", length)
+		case length > 2:
+			return nil, ExceptionNewf(TypeError, "%s expected at most 2 arguments, got %d", "items()", length)
+		}
+		sMap := self.(StringDict)
+		if str, ok := args[0].(String); ok {
+			if res, ok := sMap[string(str)]; ok {
+				return res, nil
+			}
+
+			switch length {
+			case 2:
+				return args[1], nil
+			default:
+				return None, nil
+			}
+		}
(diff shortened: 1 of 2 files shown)


py/dict.go
```diff
@@ -36,6 +36,29 @@ func init() {
 		}
 		return NewIterator(o), nil
 	}, 0, "items() -> list of D's (key, value) pairs, as 2-tuples")
+
+	StringDictType.Dict["get"] = MustNewMethod("get", func(self Object, args Tuple) (Object, error) {
+		var length = len(args)
+		if length == 0 {
```
sbinet: I believe "Effective Go" recommands to use `switch` for these kind of `if-else-if` code paths:
```go
switch {
case length == 0:
   // ...
case length > 2:
   // ...
}
...
```

py/dict.go
```diff
@@ -36,6 +36,29 @@ func init() {
...
+			return nil, ExceptionNewf(TypeError, "%s expected at least 1 arguments, got %d", "items()", length)
+		} else if length > 2 {
+			return nil, ExceptionNewf(TypeError, "%s get expected at most 2 arguments, got %d", "items()", length)
+		}
+		sMap := self.(StringDict)
+		if str, ok := args[0].(String); ok {
+			res, ok := sMap[string(str)]
+			if ok {
```
sbinet: ditto:
```go
switch res, ok := sMap[string(str)]; ok {
case ok:
    return res, nil
case length == 1:
    return None, nil
default:
    return args[1], nil
}
```
```

Comment (Sungmin-Joo) [reply py/dict.go]: @sbinet 
``` go
switch res, ok := sMap[string(str)]; ok {
case ok:
    return res, nil
case length == 1:
    return None, nil
default:
    return args[1], nil
}
```
I applied as you advised, but it caused a "panic".
So I made some modifications.

Sungmin-Joo:
@ncw 
Can I ask you to confirm my PR once again?

PR #109, opened 2019-09-30
__ne__ of dict return NotImplemented

### LINKED ISSUE #108: Dict.__ne__ doesn't return NotImplemented
Dict.__ne__ doesn't return NotImplemented.

https://github.com/go-python/gpython/blob/af17d7daed3e64897f592923022638280abc0627/py/dict.go#L167-L176

If result of `__eq__` is Not Implemented, `__ne__` should return `NotImplemented`, not `True`

### Expected result (python)
```python
>>> a = {}
>>> a.__ne__(3)
NotImplemented
```

### Actual result (gpython)
```python
>>> a = {}
>>> a.__ne__(3)
True
```

HyeockJinKim:
if type of other is not dict,
return NotImplemented

Fixes #108


HyeockJinKim:
@corona10 
Can I get a review?

py/tests/dict.py
```diff
@@ -33,4 +33,15 @@
 assert a.__contains__('hello')
 assert not a.__contains__('world')
 
+doc="__eq__, __ne__"
+a = {'a': 'b'}
+assert a.__eq__(3) != True
+assert a.__ne__(3) != False
+assert a.__ne__(3) != True
+assert a.__ne__(3) != False # This part should be changed in comparison with NotImplemented
```
ncw: Can you remove this comment please?  It is in the wrong place (it should be on the line above) and it refers to a bug which is fixed (or will be when this is merged) so will make no sense in the future.

The best way of making that sort of comment is to add a github comment to the review.

HyeockJinKim: Removed it. Would you like to review it again?

ncw:
Thank you :-)

We can think about more abstraction in a later PR if we want.

I'll merge this now.

PR #110, opened 2019-09-30
__ne__ of set return NotImplemented 

HyeockJinKim:
if type of other is not set, 
return NotImplemented


ncw:
Thank you :-)  I'll merge this now.

Issue #111, opened 2019-10-01
Implement float is_integer method


DoDaek:
## python
```
>>> a = 10.0
>>> type(a)
<class 'float'>
>>> a.is_integer()
True
```

## gpython
```
>>> a = 10.0
>>> type(a)
<class 'float'>
>>> a.is_integer()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    FIXME line of source goes here
AttributeError: "'float' has no attribute 'is_integer'"
```

`is_integer()` method does not exist.

DoDaek:
@corona10 
There are built-in functions that are not implemented. 
Can I implement them and send PR?

PR #112, opened 2019-10-02
Implement float is_integer method

DoDaek:
ISSUE : #111

### DIFF
--- a/py/float.go
+++ b/py/float.go
@@ -398,14 +398,14 @@ func (a Float) M__ge__(other Object) (Object, error) {
 func init() {
 	FloatType.Dict["is_integer"] = MustNewMethod("is_integer", func(self Object) (Object, error) {
 		if a, ok := convertToFloat(self); ok {
-			f, err := FloatAsFloat64(a)
-			if err != nil {
-				return nil, err
+			f, _ := FloatAsFloat64(a)
+			if math.Floor(f) == f {
+				return True, nil
 			}
-			b := math.Abs(f - math.Round(f))
-			return NewBool(b < math.SmallestNonzeroFloat64), nil;
+			return False, nil
 		}
-		return nil, AttributeError
+		_, e := cantConvert(self, "float")
+		return nil, e
 	}, 0, "is_integer() -> Return True if the float instance is finite with integral value, and False otherwise.")
 }
 
--- a/py/float.go
+++ b/py/float.go
@@ -398,14 +398,13 @@ func (a Float) M__ge__(other Object) (Object, error) {
 func init() {
 	FloatType.Dict["is_integer"] = MustNewMethod("is_integer", func(self Object) (Object, error) {
 		if a, ok := convertToFloat(self); ok {
-			f, _ := FloatAsFloat64(a)
-			if math.Floor(f) == f {
-				return True, nil
+			f, err := FloatAsFloat64(a)
+			if err != nil {
+				return nil, err
 			}
-			return False, nil
+			return NewBool(math.Floor(f) == f), nil
 		}
-		_, e := cantConvert(self, "float")
-		return nil, e
+		return cantConvert(self, "float")
 	}, 0, "is_integer() -> Return True if the float instance is finite with integral value, and False otherwise.")
 }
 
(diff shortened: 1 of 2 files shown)


DoDaek:
In Go: 1.9.x, there is no math.Round method 

```
github.com/go-python/gpython/py
# github.com/go-python/gpython/py
py/float.go:405:22: undefined: math.Round
The command "eval go get -v -t ./... " failed 3 times.
```

sbinet:
we could drop support for `1.9` (and `1.10`) and bump to `1.11.x`, `1.12.x` and `1.13.x` (as is done by the Go project).

sbinet:
please rebase on top of the latest `master` (6ba973e) to fix the travis failure.

py/float.go
```diff
@@ -394,6 +394,21 @@ func (a Float) M__ge__(other Object) (Object, error) {
...
+// Properties
+func init() {
+	FloatType.Dict["is_integer"] = MustNewMethod("is_integer", func(self Object) (Object, error) {
+		if a, ok := convertToFloat(self); ok {
+			f, err := FloatAsFloat64(a)
+			if err != nil {
+				return nil, err
+			}
+			b := math.Abs(f - math.Round(f))
+			return NewBool(b < math.SmallestNonzeroFloat64), nil;
```
corona10: Is there any reason to use this logic?
CPython check the float is an integer by using the floor API.
``` python
 o = (floor(x) == x) ? Py_True : Py_False;
```

py/float.go
```diff
@@ -394,6 +394,21 @@ func (a Float) M__ge__(other Object) (Object, error) {
...
+		}
+		return nil, AttributeError
```
corona10:
````
go
return nil, cantConvert(a, "float")
````
might be better

File: py/float.go (hunk 1 above)
Comment (DoDaek) [reply py/float.go]: I calculated the difference between rounded value and original value.
If the value was less than the minimum value of float64, I think that it would be an integer.

I'll check the Cpython and then fix that part by using floor API.

Comment (corona10) [reply py/float.go]: FYI, https://github.com/python/cpython/blob/279f44678c8b84a183f9eeb85e0b086228154497/Objects/floatobject.c#L853

File: py/float.go (hunk 2 above)
Comment (DoDaek) [reply py/float.go]: I'll change that part with cantConvert method.

File: py/float.go [hunk 3]
```diff
@@ -398,14 +398,14 @@ func (a Float) M__ge__(other Object) (Object, error) {
 func init() {
 	FloatType.Dict["is_integer"] = MustNewMethod("is_integer", func(self Object) (Object, error) {
 		if a, ok := convertToFloat(self); ok {
-			f, err := FloatAsFloat64(a)
-			if err != nil {
-				return nil, err
+			f, _ := FloatAsFloat64(a)
```
corona10: Please return nil, err when FloatAsFloat64 return error

py/float.go
```diff
@@ -398,14 +398,14 @@ func (a Float) M__ge__(other Object) (Object, error) {
...
+			if math.Floor(f) == f {
+				return True, nil
 			}
-			b := math.Abs(f - math.Round(f))
-			return NewBool(b < math.SmallestNonzeroFloat64), nil;
+			return False, nil
 		}
-		return nil, AttributeError
+		_, e := cantConvert(self, "float")
```
corona10: Please return cantConvert directly

py/float.go
DoDaek: Ok, I'll change that.

py/float.go
DoDaek: How do I use the second return value directly?

Sorry, I'm looking for that. But I can't find any information.

py/float.go
```diff
@@ -394,6 +394,21 @@ func (a Float) M__ge__(other Object) (Object, error) {
...
+func init() {
+	FloatType.Dict["is_integer"] = MustNewMethod("is_integer", func(self Object) (Object, error) {
+		if a, ok := convertToFloat(self); ok {
+			f, _ := FloatAsFloat64(a)
+			if math.Floor(f) == f {
+				return True, nil
+			}
+			return False, nil
+		}
+		_, e := cantConvert(self, "float")
```
sbinet:
```suggestion
		return cantConvert(self, "float")
```

sbinet: hth :)

DoDaek: Thanks for helping me out.
I couldn't think so....

PR #117, opened 2019-10-08
Implement set repr

### LINKED ISSUE #85: Implement set repr
## Expected

````python

>>> a = {1, 2, 3}
>>> a
{1, 2, 3}
````

## Actual
````python
>>> a = {1, 2, 3}
>>> a
<set instance at 0xc0000aa018>
````

### PR DESCRIPTION
**Pull request by xarus01**
Fix #85

### CODE REVIEW

File: py/tests/set.py [hunk 1]
```diff
@@ -56,6 +56,15 @@
 d = a ^ b
 assert 1 in c
 
+doc="__repr__"
+a = {1, 2, 3}
+b = a.__repr__()
```
sbinet: what about this:
```suggestion
assert "{1, 2, 3}" == a.__repr__()
```

instead?

xarus01: @sbinet Set is an unordered data structure and I don't think comparing return value of __repr__ with string will be good test case since order of elements are not guaranteed to be same.

sbinet: Right :)

builtin/testfile
```diff
@@ -0,0 +1 @@
+1,2,3,
```
corona10: Please remove thie file

go.mod
```diff
@@ -4,3 +4,5 @@ require (
 	github.com/gopherjs/gopherwasm v1.0.0 // indirect
 	github.com/peterh/liner v1.1.0
 )
+
+go 1.13
```
corona10: Should this line be added?

PR #122, opened 2019-10-20
Implement isinstance

xarus01:
#121

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -44,7 +44,7 @@ func init() {
 		// py.MustNewMethod("hex", builtin_hex, 0, hex_doc),
 		// py.MustNewMethod("id", builtin_id, 0, id_doc),
 		// py.MustNewMethod("input", builtin_input, 0, input_doc),
-		// py.MustNewMethod("isinstance", builtin_isinstance, 0, isinstance_doc),
+		py.MustNewMethod("isinstance", builtin_isinstance, 0, isinstance_doc),
 		// py.MustNewMethod("issubclass", builtin_issubclass, 0, issubclass_doc),
 		py.MustNewMethod("iter", builtin_iter, 0, iter_doc),
 		py.MustNewMethod("len", builtin_len, 0, len_doc),
@@ -825,6 +825,45 @@ Read and execute code from an object, which can be a string or a code
 object.
 The globals and locals are dictionaries, defaulting to the current
 globals and locals.  If only globals is given, locals defaults to it.`
+
+const isinstance_doc = `isinstance(obj, class_or_tuple) -> bool
+
+Return whether an object is an instance of a class or of a subclass thereof.
+
+A tuple, as in isinstance(x, (A, B, ...)), may be given as the target to
+check against. This is equivalent to isinstance(x, A) or isinstance(x, B)
+or ... etc.
+`
+
+func isinstance(obj py.Object, classOrTuple py.Object) (py.Bool, error) {
+	switch classOrTuple.(type) {
+	case py.Tuple:
+		var class_tuple = classOrTuple.(py.Tuple)
+		for idx := range class_tuple {
+			res, _ := isinstance(obj, class_tuple[idx])
+			if res {
+				return res, nil
+			}
+		}
+		return false, nil
+	default:
+		if classOrTuple.Type().ObjectType != py.TypeType {
+			return false, py.ExceptionNewf(py.TypeError, "isinstance() arg 2 must be a type or tuple of types")
+		}
+		return obj.Type() == classOrTuple, nil
+	}
+}
+
+func builtin_isinstance(self py.Object, args py.Tuple) (py.Object, error) {
+	var obj py.Object
+	var classOrTuple py.Object
+	err := py.UnpackTuple(args, nil, "isinstance", 2, 2, &obj, &classOrTuple)
+	if err != nil {
+		return nil, err
+	}
+
+	return isinstance(obj, classOrTuple)
+}
 
 const iter_doc = `iter(iterable) -> iterator
 iter(callable, sentinel) -> iterator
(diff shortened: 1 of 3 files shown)


builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+		}
+	}
+}
+
+func builtin_isinstance(self py.Object, args py.Tuple) (py.Object, error) {
+	if len(args) > 2 {
+		return nil, py.ExceptionNewf(py.TypeError, "isinstance expected 2 arguments, got %d", len(args))
+	}
+
+	return isinstance(args)
```
tim-st: When you use `err := UnpackTuple(args, nil, "isinstance", 2, 2, obj, classOrTuple)` it will check all cases and you don't have to write your own error message, which would increase binary size.

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+func isinstance(args py.Tuple) (py.Bool, error) {
+	switch args[1].(type) {
+	case py.Tuple:
+		{
+			var res py.Bool
+			var class = args[1].(py.Tuple)
+			for idx := range class {
+				var new_args []py.Object
+				new_args = append(new_args, args[0], class[idx])
+				temp, err := isinstance(new_args)
```
ncw: I didn't know isinstance recursively unpacked its arguments!

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+
+A tuple, as in isinstance(x, (A, B, ...)), may be given as the target to
+check against. This is equivalent to isinstance(x, A) or isinstance(x, B)
+or ... etc.
+`
+
+func isinstance(args py.Tuple) (py.Bool, error) {
+	switch args[1].(type) {
+	case py.Tuple:
+		{
```
ncw: Can you remove the braces here - they aren't necessary

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+				temp, err := isinstance(new_args)
+				if err != nil {
+					return false, err
+				}
+				res = res || temp
+			}
+			return res, nil
+		}
+	default:
+		{
```
ncw: Remove these braces please

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+`
+
+func isinstance(args py.Tuple) (py.Bool, error) {
+	switch args[1].(type) {
+	case py.Tuple:
+		{
+			var res py.Bool
+			var class = args[1].(py.Tuple)
+			for idx := range class {
+				var new_args []py.Object
```
ncw: This would be more efficently written

```
var new_args = []py.Object{args[0], class[idx]}
```

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+				new_args = append(new_args, args[0], class[idx])
+				temp, err := isinstance(new_args)
+				if err != nil {
+					return false, err
+				}
+				res = res || temp
```
ncw: This should say

```go
if res {
    return res
}
```

to short circuit the rest of the evaluations shouldn't it?

If you do that you can move the definition of res to here.

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+			}
+			return res, nil
```
ncw: Then this can say

```go
return false, nil
```

builtin/builtin.go
```diff
@@ -826,6 +826,50 @@ object.
...
+		}
+	default:
+		{
+			if args[0].Type() != py.TypeType {
```
ncw: You are checking this for every entry in the args - it would be better checked in builtin_isinstance I think.

builtin/builtin.go
xarus01: @tim-st I agree with your point. Used UnpackTuple function instead.

tim-st: @xarus01 I think the line `return nil, err` didn't format by your IDE, otherwise this change looks good!

vm/tests/builtin.py
```diff
@@ -65,4 +65,25 @@
...
+try:
+    isinstance(1, (A, ), "foo")
+except TypeError:
+    pass
+else:
+    assert False, "TypeError not raised"
+
+try:
+    isinstance(1, [A, "foo"])
+except TypeError:
```
corona10: https://github.com/go-python/gpython/blob/0c23b142f7b58bc925071df6ffdb04d7c112a21c/py/tests/libtest.py#L9

Please use `assertRaises` for this types of test. :)

builtin/builtin.go
```diff
@@ -826,6 +826,46 @@ object.
...
+		}
+		return obj.Type() == classOrTuple, nil
+	}
+}
+
+func builtin_isinstance(self py.Object, args py.Tuple) (py.Object, error) {
+  var obj py.Object
+  var classOrTuple py.Object
+  err := py.UnpackTuple(args, nil, "isinstance", 2, 2, &obj, &classOrTuple)
+	if err != nil {
```
corona10: gofmt?

vm/tests/builtin.py
xarus01: @corona10 Thanks! I created libtest.py to `vm/tests/` directory.

builtin/builtin.go
xarus01: @corona10 My bad.

builtin/builtin.go
```diff
@@ -826,6 +826,46 @@ object.
...
+
+Return whether an object is an instance of a class or of a subclass thereof.
+
+A tuple, as in isinstance(x, (A, B, ...)), may be given as the target to
+check against. This is equivalent to isinstance(x, A) or isinstance(x, B)
+or ... etc.
+`
+
+func isinstance(obj py.Object, classOrTuple py.Object) (py.Bool, error) {
+  print(obj.Type().Name, "\t", classOrTuple.Type().Name, "\n")
```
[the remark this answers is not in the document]
xarus01: @ncw Thanks for pointing out!

ncw:
I think this is looking good now - thank you :-)

Apologies for the delay!

I'll merge it now.

PR #123, opened 2019-11-05
builtin,vm: add implementation for builtin hex function

sbinet:


sbinet:
does anybody understand why _e.g._ `builtin_isinstance` and `builtin_hex` are marked as "not covered" while _e.g._ `builtin_sorted` is?

For the life of me, I just don't understand.

ncw:
Codecov seems to do mad things sometimes! Did you force push your branch? I think that makes it worse sometimes.

I wonder if there is a retry button for codecov?

sbinet:
no, I didn't force push. (at least, not for this PR :P)
and actually running `go test -cover` also shows the same behaviour wrt `builtin_hex` and `builtin_sorted`.

ncw:
Very strange. I can't see anything wrong with your code. Might be worth trying with different go versions to see if it is a go bug.

sbinet:
ok. figured it out.

the tests (for `builtin_hex` and `builtin_isinstance`) were in the wrong location (and actually badly written) :)

sbinet:
@ncw PTAL

PR #125, opened 2019-11-05
Initial attempt at gometalinter rules

sbinet:


sbinet:
it seems f4ab05f was dropped from the main tree.
adding it back.

PR #126, opened 2019-11-06
Implemented .pop() method for list

### LINKED ISSUE #95: Implement list pop method
Attribute errors occured using list.
​
**gpython**
```
>>> a = [1,2,3,4]
>>> a.pop()
AttributeError: "'list' has no attribute 'pop'"
```
​
**python**
```
>>> a.pop()
4
>>>a
[1,2,3]
>>>a.pop(1)
2
>>>a
[1,3]
```

xarus01:
Fixes #95.

### DIFF
--- a/py/list.go
+++ b/py/list.go
@@ -28,6 +28,22 @@ func init() {
 		return NoneType{}, nil
 	}, 0, "append(item)")
 
+	ListType.Dict["pop"] = MustNewMethod("pop", func(self Object, args Tuple) (Object, error) {
+		var index Object = Int(0)
+		listSelf := self.(*List)
+		err := UnpackTuple(args, nil, "pop", 0, 1, &index)
+		if err != nil {
+			return nil, err
+		}
+		i, err := IndexIntCheck(index, listSelf.Len())
+		if err != nil {
+			return nil, err
+		}
+		popElement := listSelf.Items[i]
+		listSelf.Items = append(listSelf.Items[:i], listSelf.Items[i+1:]...)
+		return popElement, nil
+	}, 0, "pop(index)")
+
 	ListType.Dict["extend"] = MustNewMethod("extend", func(self Object, args Tuple) (Object, error) {
 		listSelf := self.(*List)
 		if len(args) != 1 {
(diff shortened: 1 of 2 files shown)


py/list.go
```diff
@@ -28,6 +28,19 @@ func init() {
 		return NoneType{}, nil
 	}, 0, "append(item)")
 
+	ListType.Dict["pop"] = MustNewMethod("pop", func(self Object, args Tuple) (Object, error) {
+		var index Object = Int(0)
+		listSelf := self.(*List)
+		UnpackTuple(args, nil, "pop", 0, 1, &index)
```
sbinet: please add a check of the returned error.

xarus01: @sbinet My bad. Fixed it!

corona10:
@xarus01  CI is failed. Please take a look

ncw:
Here is the failure.  This is trying to run the tests under python3.4.

```
============================================================
Failures for /usr/bin/python3.4 in ./py/tests/list.py
============================================================
  File "./py/tests/list.py", line 40, in <module>
    assert a.pop() == 1
AssertionError
Original exception was:
  File "./py/tests/list.py", line 40, in <module>
    assert a.pop() == 1
AssertionError
============================================================
```

(there was more stuff I've snipped out the error which seems to be python trying to report errors via apport and failing :confused: )

py/tests/list.py
```diff
@@ -35,6 +35,13 @@
 assert repr(a) == "['a', 'b', 'c', 'd', 'e', 'f']"
 assertRaises(TypeError, lambda: [].append())
 
+doc="pop"
+a = [1,2,3,4]
+assert a.pop() == 1
+assert a.pop(2) == 4
```
tim-st: The `pop` operation removes and returns from the back, I think your idea was removing from the front.
It would be good if you create a new list for each test and compare your test against a list like `assert a.pop(-1) == elemtent; assert a = [...]`, that makes it more clear here.

Additionally the following tests would be good:
* negative index
* index out of range for non-empty list
* parameter of wrong type (non-int)

PR #133, opened 2020-06-11
expose Token, Value, TokenToName and Pos on LexToken

sneakywombat:
I am using your library to write a custom parser, and thus only depend on the lex functionality.  While writing it, I found it useful to be able to access these values.  Tossing it up here to see if you agree.  thanks for writing such a new library!

### DIFF
--- a/parser/lexer.go
+++ b/parser/lexer.go
@@ -330,6 +330,15 @@ func (lt *LexToken) String() string {
 	return fmt.Sprintf("%q (%d) = %T{%v} %d:%d", name, lt.token, lt.value, lt.value, lt.pos.Lineno, lt.pos.ColOffset)
 }
 
+// TokenToName returns the string name of a given token
+func (lt *LexToken) TokenToName() string {
+        name, ok := tokenToString[lt.token]
+        if !ok {
+                return ""
+        }
+        return name
+}
+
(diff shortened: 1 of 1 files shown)


parser/lexer.go
```diff
@@ -330,6 +330,30 @@ func (lt *LexToken) String() string {
 	return fmt.Sprintf("%q (%d) = %T{%v} %d:%d", name, lt.token, lt.value, lt.value, lt.pos.Lineno, lt.pos.ColOffset)
 }
 
+// TokenToName returns the string name of a given token
+func (lt *LexToken) TokenToName() string {
+        name, ok := tokenToString[lt.token]
```
sbinet, line 335: perhaps we could make the token implement `fmt.Stringer` instead ?

sneakywombat: sure, i'll take a look. random question, i hacked in comment support in my branch because I wanted to capture those.  any interest in supporting comment capture in gpython?  I saw that it was tossed away in lexer.go.

sbinet: IMHO, it would be useful.
but @ncw would know best.

ncw: Comment support sounds interesting. Is it part of the standard Python lexer? I don't remember

PR #137, opened 2021-02-04
Update syntax support to python 3.8.

adam-azarchs:
Use of any new python 3.8 features will not actually work correctly in the interpreter, but parsing will work.  Due to the way that gpython's unit tests work, by comparing output from the cpython interpreter to output from itself, these changes were a bit more invasive than I'd originally planned, in order to make the tests pass.


adam-azarchs:
Ah, sorry I meant to open this PR internally against our own master, before sending this to you, as it's still a work in progress and I haven't gotten the `compile` tests passing yet.

ncw:
I had a look through the work in progress - looks like you are doing all the right things :-)

    adsl -> types -> grammar -> parser -> interpreter !

Did you update the version of python used for testing? https://github.com/go-python/gpython/blob/master/bin/install-python.sh

PR #144, opened 2021-12-10
Multi-context execution (py.Context)

drew-512:
Hi Seb, it's a been a couple months but things are well here.

With this PR, I am pleased to share gpython with multi-context execution!  Check out `type Context interface` and `vm/vm_test.go` for a quick start.

My intention with this PR is get feedback from you on what you want me to alter or massage so that you are satisfied.  I am totally open if you want me to shift, rename, or organize code in ways that you would prefer, so kindly please just let me know.  I think many will be highly interested in a multi-ctx gpython (as many have asked for it as you already know).

So if you like things as they are, you can merge this PR. Otherwise, just give me a list of things you want changed and I'll resubmit a new PR with that.  Or, perhaps merge this PR as is, insert edits as you see fit, tag it into a new release and version, and then I will follow your lead and ditch my fork and we can be on our way.  :)

We are definitely using gpython here for our classified particle research project and would love to join forces with you on maintaining gpython.  I think it is highly reasonable, especially with this PR, that gpython is picked up (sponsored) since it serves the Google, Python, and Go communities quite uniquely.


drew-512:
Liking the your CI scripts! 

Added two more commits that addressed those tests not passing

sbinet:
thanks. it's quite a large PR :)
I'll try to digest it over the coming days (with probably a hiatus Tue-Wed).

modules/runtime.go
```diff
@@ -0,0 +1,196 @@
...
+	sys_mod.Globals["argv"] = py.NewListFromStrings(opts.SysArgs)
+	sys_mod.Globals["path"] = py.NewListFromStrings(opts.SysPaths)
+
+	return ctx
+}
+
+type ctx struct {
+	store *py.Store
+	opts  py.CtxOpts
+}
```
sbinet, line 196: I'd move this at the top (with `type ctx struct {...}` at the very top)
and move `resolvePath` after it's needed (after `ctx.RunFile`)

py/frame.go
```diff
@@ -26,6 +26,7 @@ type TryBlock struct {
 // A python Frame object
 type Frame struct {
 	// Back       *Frame        // previous frame, or nil
+	Ctx             Ctx
```
sbinet, line 29: please add some comment about this field.

py/function.go
```diff
@@ -26,7 +27,6 @@ type Function struct {
 	Name        string     // The __name__ attribute, a string object
 	Dict        StringDict // The __dict__ attribute, a dict or NULL
 	Weakreflist List       // List of weak references
-	Module      Object     // The __module__ attribute, can be anything
```
sbinet, line 29: naive question: why do we drop this field?
is it available now from `Ctx`?

py/import.go
```diff
@@ -78,68 +80,46 @@ var (
...
+
+	module, err := ctx.RunFile(pathParts, opts)
+	if err != nil {
+		return nil, err
 	}
-	return nil, ExceptionNewf(ImportError, "No module named '%s'", name)
+
+	return module, nil
+
+	//return nil, ExceptionNewf(ImportError, "No module named '%s'", name)
```
sbinet, line 122: please remove.

py/module.go
```diff
@@ -6,24 +6,94 @@
...
+	rt.ModuleImpls[module.ModuleInfo().Name] = module
+}
+
+func NewStore() *Store {
+	return &Store{
+		modules: make(map[string]*Module),
+	}
+}
+
+// A python Module object that has been initted for a given py.Ctx
```
sbinet, line 91:
```suggestion
// A python Module object that has been initialized for a given py.Ctx
```

py/module.go
```diff
@@ -43,45 +113,55 @@ func (m *Module) GetDict() StringDict {
...
 }
 
 // Gets a module
-func GetModule(name string) (*Module, error) {
-	m, ok := modules[name]
+func (store *Store) GetModule(name string) (*Module, error) {
+	m, ok := store.modules[name]
 	if !ok {
-		return nil, ExceptionNewf(ImportError, "Module %q not found", name)
+		return nil, ExceptionNewf(ImportError, "Module '%q' not found", name)
```
sbinet, line 157:
```suggestion
		return nil, ExceptionNewf(ImportError, "Module '%s' not found", name)
```
(`'%q'` sounds a bit redundant)

py/run.go
```diff
@@ -0,0 +1,93 @@
...
+
+type RunFlags int32
+
+const (
+	// RunOpts.FilePath is intelligently interepreted to load the appropriate pyc object (otherwise new code is generated from the implied .py file)
+	SmartCodeAcquire RunFlags = 0x01
+)
+
+// Ctx is gpython virtual environment instance ("context"), providing a mechanism
+// for multple gpython interpreters to run concurrently without restriction.
```
sbinet, line 19:
```suggestion
// for multiple gpython interpreters to run concurrently without restriction.
```

py/run.go
```diff
@@ -0,0 +1,93 @@
...
+var (
+	DefaultCtxOpts = func() CtxOpts {
+		opts := CtxOpts{
+			SysPaths: CoreSysPaths,
+		}
+		opts.SysPaths = append(opts.SysPaths, AuxSysPaths...)
+		return opts
+	}
+	NewCtx  func(opts CtxOpts) Ctx
+	Compile func(str, filename string, mode CompileMode, flags int, dont_inherit bool) (Object, error)
```
sbinet, line 92: these would need some comment.
(and where/when/why they are initialized)

py/util.go
```diff
@@ -0,0 +1,147 @@
...
+		v_float = float64(val)
+	} else if val, ok := src.(String); ok {
+		v_str = string(val)
+		intval, _ := strconv.Atoi(v_str)
+		v_int = int64(intval)
+	} else if _, ok := src.(NoneType); ok {
+		// No-op
+	} else {
+		return ErrUnsupportedObjType
+	}
```
sbinet, line 98: wouldn't this be easier with a `type-switch` ?
```go
switch v := src.(type) {
case Bool:
    if v { ... } else { ... }
case Int:
    v_int = int64(v)
    v_float = float64(v)
...
default:
    return ErrUnsupportedObjType
}
```

vm/vm_test.go
```diff
@@ -17,3 +22,72 @@ func TestVm(t *testing.T) {
...
+	workersRunning := sync.WaitGroup{}
+
+	numJobs := 35
+	fmt.Printf("Starting %d workers to process %d jobs...\n", numWorkers, numJobs)
+
+	jobPipe := make(chan int)
+	go func() {
+		for i := 1; i <= numJobs; i++ {
+			jobPipe <- i
+		}
```
sbinet, line 69:
```suggestion
		for i := 0; i < numJobs; i++ {
			jobPipe <- i+1
		}
```
(`for`-loops with "strange" arguments are strange)

sbinet (commented):
a very first pass: I see there are a bunch of files which have a `0644` -> `0755` `chmod` change.
could this be reversed?
(I don't think this brings anything to these files. or does it?)

a naive question: do we expect multiple `py.Ctx` implementations? wouldn't be easier to start with just a concrete `Ctx` in `py`?

also: there are a couple of fixes which are useful to have at some point but aren't needed for the multi-context change. could these be part of another PR?

thanks.

modules/runtime.go
```diff
@@ -0,0 +1,196 @@
+package modules
```
sbinet, line 1: please add a `LICENCE` blurb:

```
// Copyright 2021 The go-python Authors.  All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

```

(ditto for all other new files. thanks)

sbinet:
also, it would be great to have an idea of the (probably) loss of performances _wrt_ "single context" execution.

go.sum
```diff
@@ -1,8 +1,13 @@
+github.com/go-python/py v0.0.0-20150724143143-adb74cbb1b5b h1:vTYiMD/1XTeL0SwwyVG30JveBV9ZMgJEZBwDcixC/oM=
```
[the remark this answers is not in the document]
drew-512: good catch

py/function.go
drew-512: I am open to to add it back -- I just didn't see a use for it (and seemed unused before).  Now that `type Method struct` has a valid`*Module` set, module Methods now always have self set properly:

https://github.com/drew-512/gpython/blob/5ebaf66936cc1f96e34a348cf7066c0369d9f7c7/py/method.go#L236

drew-512:
> also, it would be great to have an idea of the loss (probably?) of performances _wrt_ "single context" execution.

Please check out `examples/multi-ctx.go` and tell me what you think?   What do you suggest I do from here -- I am totally open to ideas and don't have a lot of experience w/ Go's test/bench features.

Other than the `py/function.go` issue still open for resolution above and how this multi-ctx demo, are you seeing anything else my friend?

If you think I should make a concrete py.Ctx (struct) rather than an interface, I am open to that.  Is your basis for this because there is less overhead?  

I tried reverted everything to 644, sorry about there.  There were a few scripts remaining and I wasn't sure what made sense there.

Best,
Drew

drew-512:
I'm not understanding what is happening with the 644 issue.  I checkout a new copy of gpython and I see 755 everywhere, so I'm not clear what I should be doing or missed, sorry.  Please let me know what I can do, thanks.

py/util.go
```diff
@@ -0,0 +1,155 @@
+package py
```
sbinet, line 1: please add a `LICENSE` blurb.

vm/eval.go
```diff
@@ -1087,9 +1087,11 @@ func do_IMPORT_NAME(vm *Vm, namei int32) error {
 	}
 	v := vm.POP()
 	u := vm.TOP()
-	var locals py.Object = vm.frame.Locals
-	if locals == nil {
+	var locals py.Object
+	if vm.frame.Locals == nil {
 		locals = py.None
+	} else {
+		locals = vm.frame.Locals
```
sbinet, line 1094:
```suggestion
	var locals py.Object = py.None
	if vm.frame.Locals != nil {
		locals = vm.frame.Locals
```

vm/eval.go
```diff
@@ -2033,7 +2036,14 @@ func tooManyPositional(co *py.Code, given, defcount int, fastlocals []py.Object)
 		chooseString(given == 1 && kwonly_given == 0, "was", "were"))
 }
 
-func EvalCodeEx(co *py.Code, globals, locals py.StringDict, args []py.Object, kws py.StringDict, defs []py.Object, kwdefs py.StringDict, closure py.Tuple) (retval py.Object, err error) {
+// Run a new virtual machine on a Code object
```
sbinet, line 2039:
```suggestion
// EvalCode runs a new virtual machine on a Code object.
```

examples/multi-ctx/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"fmt"
+	"log"
+	"runtime"
+	"strings"
+	"sync"
+	"time"
+
+	_ "github.com/go-python/gpython/modules"
```
sbinet, line 11: as this is an example, perhaps comment on why this is needed?

sbinet:
could you rebase off the latest `master`?

(there's a little conflict in `go.mod` as I have bumped the minimum Go version to `1.16`)

examples/multi-ctx/main.go
drew-512: great suggestion

modules/runtime.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+
+			out.Code, err = py.Compile(string(pySrc), fpath, py.ExecMode, 0, true)
+			if err != nil {
+				return false, err
+			}
+			out.SrcPathname = fpath
+		} else if ext == ".pyc" {
+			var file *os.File
+			file, err = os.Open(fpath)
```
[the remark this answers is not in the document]
drew-512: good catch

examples/multi-ctx/main.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/go-python/gpython/py"
+)
+
+func main() {
+
+	// The total job count implies a fixed amount of work.
```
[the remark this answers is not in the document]
drew-512: lol oops

drew-512:
Hi @sbinet , excited to be supporting you on all this and thanks for the catches and suggestions.

As I mentioned in the other thread, I'll update the README for your review in the coming days.

Also, one idea I wanted to explore was renaming `py.Ctx` to `py.Context`?  Either seem ok or perhaps you have another.  I defer to your experience friend.

Drew

drew-512:
Good suggestions in your opening comment and will do a revision.

sbinet:
`s/py.Ctx/py.Context/g` is fine by me.

sbinet (commented):
getting there.
thanks for sticking w/ me.

(if I am not mistaken, the last remaining thing to do is setup a little "single-ctx" benchmark to assess the performance drop coming w/ contexts, in order to have a ballpark number in mind)

drew-512:
Love all the CI stuff you're adding, awesome!

The items I'm working on in the coming days are a README rev, a well documented embedded example (huge for newcomers), and the single vs multi ctx benchmark as you suggested.    I think gpython as a viable, reliable embedded interpreter solution has very high potential.  Something as potent, easy, and fast to embed as Lua, but in general, devs don't ever want to choose that b/c they don't want to tell their users to have to learn `{{$esoteric_embedded_language}}`.   My vision is that `RunFile()`, multi-context, and this embedding example should be an inviting and positive signs for those evaluating gpython for the first time.

I've heard many big cats often quip "the second most popular language in the world", reflecting that when it comes to non-developers (finance, science), Python has the best learning curve and those users typically already have some python experience.  Also, many of those types want to grow their python skills b/c they see it show up all over their career.   So I suspect this could be a big theme, now we are in world awash in interpreters that are basically only interesting to professional developers.   The gpython multi-context additions for this physics research group I mentioned are working out very well and seem to "complete" gpython for real world Go applications.   

I would love to remind you and @ncw what a huge freaking achievement running any python 3 in Go is, and that gpython's potential is not to be underestimated as a leading Go embedding option.  Or perhaps I see no reason why Google would not want to donate a summer's worth of resources to gpython if that is something beneficial to gpython.

Drew

modules/runtime.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	if err != nil {
+		return py.CompileOut{}, err
+	}
+
+	return out, nil
+}
+
+func (ctx *context) PushBusy() error {
```
sbinet, line 184: it seems to `PushBusy` is only used inside the `modules` package.
so we can probably just not export it.
```suggestion
func (ctx *context) pushBusy() error {
```

modules/runtime.go
```diff
@@ -0,0 +1,283 @@
...
+	if ctx.closed {
+		return py.ExceptionNewf(py.RuntimeError, "Context closed")
+	}
+	ctx.running.Add(1)
+	return nil
+}
+
+func (ctx *context) PopBusy() {
```
sbinet, line 192: ditto:

```suggestion
func (ctx *context) popBusy() {
```

py/run.go
```diff
@@ -0,0 +1,175 @@
...
+	// RunCode is a lower-level invocation to execute the given py.Code.
+	RunCode(code *Code, globals, locals StringDict, closure Tuple) (result Object, err error)
+
+	// Returns the named module for this context (or an error if not found)
+	GetModule(moduleName string) (*Module, error)
+
+	// Gereric access to this context's modules / state.
+	Store() *ModuleStore
+
+	// Close signals that is context is about to go out of scope and any internal resources should be released.
```
sbinet, line 42:
```suggestion
	// Close signals this context is about to go out of scope and any internal resources should be released.
```

py/run.go
```diff
@@ -0,0 +1,175 @@
...
+	// Operations on a py.Context that have closed will generally result in an error.
+	Close()
```
sbinet, line 44: any reason why `Close()` wouldn't return an `error`?

modules/runtime.go
```diff
@@ -0,0 +1,283 @@
...
+	// Assign the base-level py.Context creation function while also preventing an import cycle.
+	py.NewContext = NewContext
+}
+
+// context implements py.Context
+type context struct {
+	store     *py.ModuleStore
+	opts      py.ContextOpts
+	closeOnce sync.Once
+	closing   bool
```
sbinet, line 36: is `closing` that useful?
it doesn't seem to be really used as of yet.

sbinet (commented):
thanks for the update.

but, before going further, could we try to address the "single-thread before/after-context" benchmarking issue?

*then* I'd like to merge this PR in, so as to reduce the amount of reviewing work.
thanks a lot.

drew-512:
Indeed!  :D  

Good suggestions, btw.  

Commits coming in by the end of this weekend is a big documented embedding example, readme edits, and that benchmark.

drew-512:
I'll switch to do and commit that benchmark first -- just had been putting that off, waaah.

drew-512:
@sbinet for the pre-context vs with-context benchmarks you raised, I'm not clear on what the path is.  I am getting that you want a PR that does bench X on gpython before `py.Context` and then see how the test performs with the `py.Context` applied.  

Questions:

1) Is the intention/purpose to just see if there's a load penalty of an extra `interface{}` traveling around everywhere?   I mean, I can't see how it would be anything but negligible, especially when you look at how much GC load there is from any kind of execution.  Really, the only change is that instead of going to a global module lookup, the lookup occurs on a context-instanced map which of course is lateral.

2) Because the`py.Context` is just another field in the `VM` struct, I'm not even sure how any python code can/would test any performance hit, other than, say, comparing the pystone of `pi_chudnovsky_bs.py` before and after the `py.Context` adds.  In your prev post, you mentioned a "local" variable edit -- do you mean embedded callbacks or editing a module local?  

Thanks.

modules/runtime.go
drew-512: The intent is that Close() can be use reliably and easily, so multiple concurrent Close() support seems important.  And when it comes at no allocation cost, that makes sense to me.

drew-512:
Hi everyone!  

At this point I am ceasing (new) work on my fork and awaiting guidance on anything further.  I am happy to take feedback and @sbinet please let me know on my questions above regarding bench comparison.  In summary: analysis of the `py.Context` shows, at most, the penalty can't even be measured.

Aside from readme or embedded example edits, I recommend we merge and celebrate this exciting step forward.  A lot of people will find this multi-context capability exciting and recognize what it unlocks!

drew-512:
Big thanks to everyone has helped make gpython happen over the years -- though it has been some time!

@ncw it would be great to get any feedback or your blessing on any of the code or the README.  

Otherwise, @sbinet has been making great suggestions and I think we're in great shape to almost ready to merge!   🎉

sbinet (commented):
last nitpicks.

I'd prefer to keep:
- 7d2a8d9a28639565862ca5ff915f3697d294e6cb (print kwargs)
- ba7db80d09226467ec635d00dcc182692302bb99 (kwargs)
- 61e42c8b6eddecce05958f913a231df27e2650dc (embedding)
out of this PR (and have those in dedicated PRs)

also, it would be great if you could create a dedicated branch for these PRs (easier on the review process).

could you send another PR against [go-python/licence](https://github.com/go-python/license) adding yourself to the `AUTHORS` and/or `CONTRIBUTORS` file?
(then, once the `multi-context` PR is validated, I'll merge the licence one and the `multi-context` one)

thanks.

PR #154, opened 2022-01-28
added utils.go

drew-512:
added commonly used py utility funcs

### DIFF
--- a/py/util.go
+++ b/py/util.go
@@ -1,0 +1,206 @@
+// Copyright 2022 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package py
+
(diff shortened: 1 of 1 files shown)


drew-512:
Hi @sbinet 

I'm taking a fresh stab at this with a fresh fork for our sanity.  :)

Once these PRs are in, I'll PR the core `py.Context` adds, then then PR the multi-context chudnovsky test and the embedding example (3 PRs). 

It will be easier if we merge the 153, 154, 155 so I can rebase and do the next 3,  let me know if I'm missing anything.

sbinet:
IIUC the high-level functions exposed here are later used by the "embedded" example, right?
let's incorporate this commit (possibly still as a separate commit) in the PR that actually use it?
WDYT?

PR #155, opened 2022-02-02
Add py.Context

drew-512:
Hi @sbinet 

So this PR has all the `py.Context` stuff wrapped into it minus:
   - examples/embedded
   - examples/multi-context
   - py/util.go (consumed by examples/embedded)
   - README.md edits

On the sync.CloseSync issue you raised in the prev thread, the intention is to ensure that multiple holders of a `py.Context` can all call `Close()` without any (future) consequences.  IMO, this is congruent with multiple holders of a py.Context (and how it's annoying when ya have to ensure that only one Close() is called).  People see that close code and they can feel comforted it's an interface ready for the real world.  And it basically comes at no cost or allocations.

Otherwise, lmk what you'd like to add/remove to this, otherwise I'll wait for this merge and then I'll immediately submit a PR with the above rolled into a final PR.


drew-512:
dammit, working too late and committed it all by accident.  going to do a new PR

drew-512:
dammit, working too late and committed it all by accident. going to do a new PR

PR #169, opened 2022-03-16
Add OS module

glaukiol1:
# OS Module

I added some of the basic functions/globals of the OS module, these include;

- Functions
    - os.getcwd()
    - os.chdir()
    - os.getenv()
    - os.getpid()
    - os.putenv()
    - os.unsetenv()
- Globals
    - os.error
    - os.environ

This is my first contribution to gpython, so please give me feedback and other things I should add.
The first two commits were tests i made to figure out how to create a module; I was going to make a JSON module, but then I saw that gpython didnt have an OS module, and I thought that was more important than a JSON module at this time.

Thanks!


os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+		Globals: globals,
+	})
+}
+
+// gets and parses all process enviroment variables
+func getEnvVariables() py.StringDict {
+	env_variables := os.Environ() // this gets all the enviroment variables
+	dict := py.NewStringDict()
+	for _, evar := range env_variables {
+		key_value := strings.Split(evar, "=") // returns a []string containing [key,value]
```
raff, line 50: Here you should probably use `strings.SplitN(evar, "=", 2)` in case there are values containing `=`

os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, py.ExceptionNewf(py.OSError, "Unable to get current working directory.")
+	}
+	return py.String(dir), nil
+}
+
+// change current working directory
+func chdir(self py.Object, args py.Tuple) (py.Object, error) {
+	if len(args) == 0 || len(args) > 1 {
+		return nil, py.ExceptionNewf(py.TypeError, "One argument required")
+	} else {
```
raff, line 70: If you are doing `if ... { return }` don't use else. It's more idomatic (and it's in the Zen of Go: https://the-zen-of-go.netlify.app/ :)

os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, py.ExceptionNewf(py.TypeError, "Expected 1 argument of type string")
+	}
+}
+
+// os._exit() immediate program termination; unline sys.exit(), which raises a SystemExit, this function will termninate the program immediately.
+func _exit(self py.Object, args py.Tuple) (py.Object, error) { // can never return
+	if len(args) == 0 {
+		os.Exit(0)
+		return nil, nil
```
raff, line 181: since os.Exit doesn't return you can avoid all the `return nil, nil` and have just one at the end of this method.

os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+			cargs = strings.Split(carg, " ")
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "Expected 1 or more arguments all of type string")
+		}
+		command := exec.Command(cargs[0])
+		outb, err := command.Output()
+		if err != nil {
+			return nil, py.ExceptionNewf(py.OSError, err.Error())
+		}
+		fmt.Println(string(outb))
```
raff, line 221: this should use the current gphyton Stdout (that may not be os.stdout)

os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+			if err != nil {
+				return nil, py.ExceptionNewf(py.TypeError, "Unable to parse string") // this will never execute
+			}
+			carg := strings.ReplaceAll(_carg, "'", "") // required
+
+			cargs = strings.Split(carg, " ")
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "Expected 1 or more arguments all of type string")
+		}
+		command := exec.Command(cargs[0])
```
raff, line 216: Not sure of why you are passing only `cargs[0]` instead of `exec.Command(cargs[0], cargs[1:]...)`

os/os.module.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return py.String(dir), nil
+}
+
+// change current working directory
+func chdir(self py.Object, args py.Tuple) (py.Object, error) {
+	if len(args) == 0 || len(args) > 1 {
+		return nil, py.ExceptionNewf(py.TypeError, "One argument required")
+	} else {
+		if reflect.TypeOf(args[0]).String() == "py.String" {
```
raff, line 71: I wonder if would be better/faster to use type assertion here instead of reflection:

```
  if _, ok := args[0].(py.String); ok {
  }
```

raff (commented):
Looks pretty good to me, but I would clean it up and make it more idomatic (I have added some notes).

Also, if possible the errors should be as close as possible to the error reported by CPython (I think the types are correct but the messages are different, unless you did it right but the messages have changed between Python 3.4 and Python 3.9, that is what I tried).

glaukiol1:
Hi,
I have added more commits resolving most of your comments. I have two questions;
- How can i get access to gpythons stdout?
- Is there a place where I can corresponding messages for error types?

Thanks!

raff:
Thanks for the updates.

Regarding stdin, see how I did it for the "print" builtin: 
https://github.com/go-python/gpython/blob/master/builtin/builtin.go#L192-L196 and https://github.com/go-python/gpython/blob/master/builtin/builtin.go#L205-L219

Maybe it would be worth extracting a method that can be used to print to stdout, or you can `py.Call` `builtins.print`

raff:
Regarding the error messages as far as I know there isn't an easy way to get the messages from CPython, the best way may be to write a test that you can run through both Cpython and gpython and compare (I think that's what the pytest stuff is supposed to do, with the files in the "tests" folder for each module.

glaukiol1:
Hi,
I think i fixed everything.
I added a `Println` function in `py.util.py`, it prints to the standard gpython output. Anything else i need to do? When is this PR going to be merged?

Thanks

os/os.module.go
```diff
@@ -0,0 +1,241 @@
+package os
+
+// Copyright 2022 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
sbinet, line 5: please move above `package os` (and keeping an empty line between the 2)

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			Doc:  "Miscellaneous operating system interfaces",
+		},
+		Methods: methods,
+		Globals: globals,
+	})
+}
+
+// gets and parses all process enviroment variables
+func getEnvVariables() py.StringDict {
+	env_variables := os.Environ() // this gets all the enviroment variables
```
sbinet, line 47:
```suggestion
	vs := os.Environ()
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	dict := py.NewStringDict()
+	for _, evar := range env_variables {
+		key_value := strings.SplitN(evar, "=", 2) // returns a []string containing [key,value]
```
sbinet, line 50:
```suggestion
		kv := strings.SplitN(evar, "=", 2)
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			Name: "os",
+			Doc:  "Miscellaneous operating system interfaces",
+		},
+		Methods: methods,
+		Globals: globals,
+	})
+}
+
+// gets and parses all process enviroment variables
+func getEnvVariables() py.StringDict {
```
sbinet, line 46:
```suggestion
// getEnvVariables returns the dictionary of environment variables.
func getEnvVariables() py.StringDict {
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, evar := range env_variables {
+		key_value := strings.SplitN(evar, "=", 2) // returns a []string containing [key,value]
+		dict.M__setitem__(py.String(key_value[0]), py.String(key_value[1]))
+	}
+
+	return dict
+}
+
+// gets the current working directory
+func getCwd(self py.Object, args py.Tuple) (py.Object, error) {
```
sbinet, line 58:
```suggestion
// getCwd returns the current working directory.
func getCwd(self py.Object, args py.Tuple) (py.Object, error) {
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, py.ExceptionNewf(py.OSError, "Unable to get current working directory.")
+	}
+	return py.String(dir), nil
+}
+
+// gets the current working directory in a byte list
+func getCwdb(self py.Object, args py.Tuple) (py.Object, error) {
```
sbinet, line 67:
```suggestion
// getCwdb returns the current working directory as a byte list.
func getCwdb(self py.Object, args py.Tuple) (py.Object, error) {
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, py.ExceptionNewf(py.OSError, "Unable to get current working directory.")
+	}
+	return py.Bytes([]byte(dir)), nil
+}
+
+// change current working directory
+func chdir(self py.Object, args py.Tuple) (py.Object, error) {
```
sbinet, line 76:
```suggestion
// chdir changes the current working directory to the provided path.
func chdir(self py.Object, args py.Tuple) (py.Object, error) {
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	if len(args) == 0 {
+		return nil, py.ExceptionNewf(py.TypeError, "Missing required argument 'path' (pos 1)")
+	}
+	if objectIsString(args[0]) {
```
sbinet, line 80: following Go good practices, we should inverse the logic:
```suggestion
	if !objectIsString(args[0]) {
	    return nil, py.ExceptionNewf(py.TypeError, "str expected, not "+args[0].Type().Name)
	}
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			return nil, py.ExceptionNewf(py.NotADirectoryError, "Couldn't change cwd; "+err.Error())
+		}
+		return py.None, nil
+	}
+	return nil, py.ExceptionNewf(py.TypeError, "str expected, not "+args[0].Type().Name)
+}
+
+// get a enviroment variable by key
+// Return the value of the environment variable key if it exists, or default if it doesn’t. key, default and the result are str.
+func getenv(self py.Object, args py.Tuple) (py.Object, error) {
```
sbinet, line 97:
```suggestion
// getenv returns the value of the environment variable key.
// If no such environment variable exists and a default value was provided, that value is returned.
func getenv(self py.Object, args py.Tuple) (py.Object, error) {
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+	var key py.Object      // key to search for
+	var default_ py.Object // return when not found
+	var err error          // error obj
```
sbinet, line 100:
```suggestion
	var (
		key py.Object
		dflt = py.None
		err error
	)
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if len(args) == 1 {
```
sbinet, line 102: please use a `switch` statement instead:
```go
switch len(args) {
case 1:
...
case 2:
...
default:
}
```

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			key = args[0]
+			default_ = args[1]
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "str expected (pos 2), not"+args[1].Type().Name)
+		}
+	} else {
+		return nil, py.ExceptionNewf(py.TypeError, "missing one required argument: 'name:str'")
+	}
+	var res py.Object // hold the result value
+	res, err = getEnvVariables().M__getitem__(key)
```
sbinet, line 125: creating the full `py.Dict` of environment variables, just to pick one value sounds a bit expensive...
couldn't we be a bit cleverer?
we have tested above that the provided arguments are `str`s, so we could just use the Go `os.Getenv`/`os.LookupEnv` functions.

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		default_ = py.None
+	} else if len(args) == 2 {
+		if objectIsString(args[0]) {
+			key = args[0]
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "str expected (pos 1), not "+args[0].Type().Name)
+		}
+		if objectIsString(args[1]) {
+			key = args[0]
```
sbinet, line 116: why do we reassign the same value?

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+
+// get a enviroment variable by key
+// Return the value of the environment variable key if it exists, or default if it doesn’t. key, default and the result are str.
+func getenv(self py.Object, args py.Tuple) (py.Object, error) {
+	var key py.Object      // key to search for
+	var default_ py.Object // return when not found
+	var err error          // error obj
+
+	if len(args) == 1 {
+		if objectIsString(args[0]) {
```
sbinet, line 103: better inverse the logic and return early.

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			key = args[0]
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "str expected (pos 1), not "+args[0].Type().Name)
+		}
+		default_ = py.None
+	} else if len(args) == 2 {
+		if objectIsString(args[0]) {
```
sbinet, line 110: better inverse the logic and return early.

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+			key = args[0]
+		} else {
+			return nil, py.ExceptionNewf(py.TypeError, "str expected (pos 1), not "+args[0].Type().Name)
+		}
+		if objectIsString(args[1]) {
```
sbinet, line 115: better inverse the logic and return early.

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		return py.None, nil
+	}
+	return nil, py.ExceptionNewf(py.TypeError, "missing required arguments: 'key:str' and 'value:str'")
+}
+
+// Unset (delete) the environment variable named key.
+func unsetenv(self py.Object, args py.Tuple) (py.Object, error) {
+	if len(args) == 1 && objectIsString(args[0]) {
+		if reflect.TypeOf(args[0]).String() == "py.String" {
```
sbinet, line 163: why do this when we've already tested for `objectIsString` ?

os/os.module.go
```diff
@@ -0,0 +1,241 @@
...
+		_k, err := py.ReprAsString(args[0])
+		if err != nil {
+			return nil, py.ExceptionNewf(py.TypeError, "Unable to parse string") // this will never execute
+		}
+		key := strings.ReplaceAll(_k, "'", "") // required
+		_v, err := py.ReprAsString(args[1])
+		if err != nil {
+			return nil, py.ExceptionNewf(py.TypeError, "Unable to parse string") // this will never execute
+		}
+		value := strings.ReplaceAll(_v, "'", "")
```
sbinet, line 149: this is perhaps a bit too naive an operation.

`export FOO="Bar's"`
is a valid environment variable.

py/util.go
```diff
@@ -204,3 +205,35 @@ func loadValue(src Object, data interface{}) error {
 	}
 	return nil
 }
+
+// print to gpythons stdout neatly
+func Println(self Object, args ...string) (int, error) {
```
sbinet, line 210: perhaps, instead of `self Object`, directly require `self` to be `*Module`, or the correct `Context`?
or, to follow Go's lead, require a "file-like" object? (but admittedly, that function should be called `Fprintln`...)

sbinet (commented):
hi,

thanks for the PR.
please find below a first run of comments and suggestions.

(I'll try to come back at it in the coming days, but I am a bit under the weather this days...)

glaukiol1:
Hi @sbinet,

Thanks for your review. I have resolved all of them, but the last one, specifically this one,
> perhaps, instead of `self Object`, directly require `self` to be `*Module`, or the correct `Context`? or, to follow Go's lead, require a "file-like" object? (but admittedly, that function should be called `Fprintln`...)

I think that for simplicity for calling it in embedded module functions, it is very simple to pass the `self py.Object` argument than to get a `*Module` or `Context`.

Any other things I need to implement? And when will this PR be ready to merge? It is getting quite long... 54 commits.

Thanks!

sbinet:
ok.
but then I am not sure which version of CPython the proposed behaviour is mimicking.
on my ArchLinux machine that runs CPython-3.10.2 *and* in a Docker container running CPython-3.4.10:

```
$> docker run --rm -it python:3.4
Python 3.4.10 (default, Mar 20 2019, 00:50:15) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print("hello\n","world")
hello
 world
```

with which OS/CPython-version are you seeing this behaviour?

glaukiol1:
Hello,
I'm running Python 3.9.8 on darwin.

@sbinet

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		Methods: methods,
+		Globals: globals,
+	})
+}
+
+// getEnvVariables returns the dictionary of environment variables.
+func getEnvVariables() py.StringDict {
+	vs := os.Environ()
+	dict := py.NewStringDict()
```
sbinet, line 117:
```suggestion
	dict := py.NewStringDictSized(len(vs))
```

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+	return py.String(dir), nil
+}
+
+// getCwdb returns the current working directory as a byte list.
+func getCwdb(self py.Object, args py.Tuple) (py.Object, error) {
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, py.ExceptionNewf(py.OSError, "Unable to get current working directory.")
+	}
+	return py.Bytes([]byte(dir)), nil
```
sbinet, line 141:
```suggestion
	return py.Bytes(dir), nil
```

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+func system(self py.Object, args py.Tuple) (py.Object, error) {
+	if len(args) != 1 {
+		return nil, py.ExceptionNewf(py.TypeError, "missing one required argument: 'command:str'")
+	}
+	var cargs []string
+	_carg, ok := args[0].(py.String)
+	if !ok {
+		return nil, py.ExceptionNewf(py.TypeError, "str expected (pos 1), not"+args[0].Type().Name)
+	}
+	carg := strings.ReplaceAll(string(_carg), "'", "\"")
```
sbinet, line 251: this isn't necessary if we use `/bin/sh -c ...`.
the shell will take care of this.

IIANM, it boils down to:
```suggestion
	arg, ok := args[0].(py.String)
	if !ok { ... }
```

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+
+	var command *exec.Cmd
+	if runtime.GOOS != "windows" {
+		command = exec.Command("/bin/sh", "-c", carg)
+	} else {
+		command = exec.Command(cargs[0], cargs[1:]...)
+	}
```
sbinet, line 258: here also, this could be handled by a dedicated `func`, guarded by `//go:build` tags.

_wrt_ the "Windows" handling: are you sure that's how CPython does it?
(I haven't checked the sources of CPython to be honest, but I'd assume it would be based on invoking `cmd.exe`)

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+	return py.Int(0), nil
+}
+
+// https://github.com/go-python/gpython/pull/169#discussion_r828767552 - type assertion
+func objectIsString(arg py.Object) bool {
+	if _, ok := arg.(py.String); ok {
+		return true
+	}
+	return false
+}
```
sbinet, line 277: I believe this could go away.

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return py.None, nil
+}
+
+// os._exit() immediate program termination; unlike sys.exit(), which raises a SystemExit, this function will termninate the program immediately.
+func _exit(self py.Object, args py.Tuple) (py.Object, error) { // can never return
+	if len(args) == 0 {
+		os.Exit(0)
+	}
+	if len(args) == 1 && objectIsInt(args[0]) {
```
sbinet, line 226:
```suggestion
	arg, ok := args[0].(py.Int)
	if !ok { ...  handle error ... }
```

because:
```py
>>> os._exit("foo")
TypeError: 'str' object cannot be interpreted as an integer
```

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+		_ec, err := py.GetInt(args[0])
+		if err != nil {
+			os.Exit(1)
+		}
+		exit_code, err := _ec.GoInt()
+		if err != nil {
+			os.Exit(1)
+		}
```
sbinet, line 234: not sure about this: it boils down to swallow silently that error. (error which would indicate some fundamental flaw in the assumptions about `int` and `int64`.)

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return false
+}
+
+func objectIsInt(arg py.Object) bool {
+	if _, ok := arg.(py.Int); ok {
+		return true
+	}
+	return false
+}
```
sbinet, line 284: I believe this could go away as well.

sbinet (commented):
it's taking shape nicely.
but, please refrain from adding new entries/vars/funcs to the module: let's try to finish and polish what we have now :)

one thing I am not completely sure about is whether we have correct handling of kwargs in `gpython`.
IIUC, `getenv` and `getenvb` do use kwargs/positional-args:
```python
>>> os.getenv(default="foo",key="bar")
'foo'
```

there's probably the need to use `py.UnpackTuple` here.
this also probably means we need proper handling of `kwargs`.
(this may be left for a later PR.)

@drew-512 , @ncw, @corona10 : do you remember whether we had implemented kwargs?

os/os.module.go
```diff
@@ -0,0 +1,284 @@
...
+	return py.String(":")
+}
+
+// getlinesep returns the line seperator (newline character). This is '\r\n' for windows and '\n' for posix
+func getlinesep() py.String {
+	if runtime.GOOS == "windows" {
+		return py.String("\r\n")
+	}
+	return py.String("\n")
+}
```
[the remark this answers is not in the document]
glaukiol1: @sbinet

I don't see how I could implement this in go:build constants, can you clarify or give me an example?

glaukiol1:
@sbinet 

I resolved all of your comments except the first one, which I left a comment on. Any other comments?

os/os.module.go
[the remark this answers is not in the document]
sbinet: you'd have to create 3 files:

```go
// os_posix.go

//go:build !windows && !android

package os

const (
    pathsep = ":"
    osname = "posix"
)

func system_cmd(arg string) *exec.Command { ... }
```

```go
// file os_windows.go

//go:build windows

package os

const (
    pathsep = "/"
    osname = "posix"
)

func system_cmd(arg string) *exec.Command { ... }
```

```go
// file os_android.go

//go:build android

package os

const (
    pathsep = "/"
    osname = "java"
)

func system_cmd(arg string) *exec.Command { ... }
```

note that `func getset()` is just the value of Go's `os.PathSeparator`, and
`func getdevnull()`  is `os.DevNull`

alternatively, one could just do this at runtime (using `runtime.GOOS`) but only once, inside `func init()`.

sbinet:
apologies for dropping the ball.
I'll come back on this one shortly.

sbinet:
(could you rebase off the latest `master` ?)

glaukiol1:
I'll do it soon. Im a bit busy this week though. I'll try my best

Thanks

sbinet:
_ping ?_

sbinet (approved):
LGTM.

could you send yet another PR against the [go-python/license](https://github.com/go-python/license), adding yourself to the `AUTHORS` and/or `CONTRIBUTORS` files?

then I'll merge this PR in.

thanks a lot for sticking with us.

sbinet:
(don't worry about the merge conflicts. I'll handle them)

drew-512:
Love it!  Awesome work, gents!

sbinet:
_pong?_

sbinet:
needs https://github.com/go-python/license/pull/18

sbinet:
done :)

PR #171, opened 2022-03-21
Adding missing __len__ method to dict objects

kellrott:
Old behavior:

len({"a" : 1, "b" : 2})
```
TypeError: "object of type 'dict' has no len()"
```

New behavior returns the number of elements in the dict


sbinet (approved):
LGTM

Issue #166, opened 2022-02-16
Upcoming gpython worklist


drew-512:
Below is my upcoming worklist I plan to take on.  @sbinet @ncw, please lmk what sized bites you want me to PR these in.  Small pieces, or perhaps separated by area (tests vs package py vs examples), or separated by items:

1. Add stack limit value to `py.ContextOpts` ([Issue 143](https://github.com/go-python/gpython/issues/143))
2. Additions to `py.utils`, including funcs that use `reflect` to auto-populate Go structs from a given py.Object.
3. Add `exit()` ([Issue 140](https://github.com/go-python/gpython/issues/140))
4. Add a shared set of golden util funcs to `pytest/pytest.go` so the same 30 boilerplate lines don't have to appear for every golden test.
5. Enhance `examples/multi-context' into a golden test.
6. For cleanliness and organization, I move all the built-in modules into the `modules` dir.  In a future where `gpython` gets traction as an embedded interpreter and more and more modules appear, it makes sense to me that they show up the modules dir, not the project root dir.  This is basically just a bunch of dirs moving and a few `import` lines changing.  I would be happy to do this first or as soon as makes sense.  
7. Introduce module os and a few commonly used functions -- e.g. `os.system()`
    - This is a security concern (e.g. python web demo).  Thinking I add flags to `py.ContextOpts` that correspond to  new `py.Method` flags which are checked when a `py.Module` is populated during a `py.Context` module import (instantiation).
8. Add support for `.format` using [pyfmt](https://github.com/slongfield/pyfmt) ([Issue 116](https://github.com/go-python/gpython/issues/116))
    - Before doing this, I would appreciate sign-off on `pyfmt` so I don't start investing time into that and then we don't want to add that as a dependency.   
    - I left a [note to the dev](https://github.com/slongfield/pyfmt/issues/15)

Also please lmk what order I should do these in if that's something that matters to people.   Otherwise, I listed them in the order that makes the most sense to me.  Again please lmk how the PRs should show up as I'm new to PR styles. 

Drew

sbinet:
the work plan SGTM.

as for the `pyfmt` work, I am a bit worried that there might be some impedance mismatch: `pyfmt` seems to have been written to bring the `.format` facilities to Go, which may not be completely the same than implementing the `.format` facilities in Go for Python. also, the main dev seem to have moved one and we may not get fixes upstream.
[...]

PR #174, opened 2022-03-23
all: move modules to stdlib

sbinet:
Updates #166


sbinet:
@drew-512 
in the end, I jumped the gun and moved the gpython stdlib modules under `stdlib`.

drew-512:
awesome!

PR #191, opened 2022-07-12
2022 examples tests cleanup

drew-512:
enhancements to package pytest and embedded golden cleanup


drew-512:
Soon to follow after this branch is a PR that does a golden test for the multi-context π example...

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
 	err = out.Close()
 	if err != nil {
-		t.Fatalf("could not close stdout/stderr: %+v", err)
+		return fmt.Errorf("could not close output file: %+v", err)
 	}
 
 	got, err := os.ReadFile(out.Name())
 	if err != nil {
-		t.Fatalf("could not read script output: %+v", err)
+		return fmt.Errorf("could not read script output file: %+v", err)
```
sbinet, line 265:
```suggestion
		return fmt.Errorf("could not read script output file: %w", err)
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
 	}
 
-	ref := fname[:len(fname)-len(".py")] + "_golden.txt"
-	want, err := os.ReadFile(ref)
+	if RegenGoldFiles {
+		err := os.WriteFile(task.GoldFile, got, 0644)
+		if err != nil {
+			return fmt.Errorf("could not write golden output %q: %+v", task.GoldFile, err)
```
sbinet, line 271:
```suggestion
			return fmt.Errorf("could not write golden output %q: %w", task.GoldFile, err)
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+		}
+	}
+
+	want, err := os.ReadFile(task.GoldFile)
 	if err != nil {
-		t.Fatalf("could not read golden output %q: %+v", ref, err)
+		return fmt.Errorf("could not read golden output %q: %+v", task.GoldFile, err)
```
sbinet, line 277:
```suggestion
		return fmt.Errorf("could not read golden output %q: %w", task.GoldFile, err)
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+}
+
+var (
+	taskCounter int32
+
+	// RegenGoldFiles will cause RunTestTasks() and RunScript() to overwrite their output as the "golden" file (rather than compare against it)
+	RegenGoldFiles bool
+
+	// GoldFileSuffix is the default expected suffix for a "golden" output file
+	GoldFileSuffix = "_golden.txt"
```
sbinet, line 183: do we really need this to be exposed and configurable?

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+		}
+		tasks = append(tasks, task)
+	}
+}
+
+var (
+	taskCounter int32
+
+	// RegenGoldFiles will cause RunTestTasks() and RunScript() to overwrite their output as the "golden" file (rather than compare against it)
+	RegenGoldFiles bool
```
sbinet, line 180: what about making this a `flag.Bool(..., false, ...)` ?
like we do in:
https://github.com/gonum/plot/blob/master/cmpimg/checkplot.go#L19

as you can see, in there we are a bit clever: when provided with `"output.png"` as an output, we infer the golden file to be `"output_golden.png"`.
we could get smart like that.

examples/embedding/main_test.go
```diff
@@ -5,58 +5,12 @@
...
-	}
-
-	want, err := os.ReadFile(fname)
-	if err != nil {
-		t.Fatalf("could not read golden file: %+v", err)
-	}
-	if !bytes.Equal(got, want) {
-		t.Fatalf("stdout differ:\ngot:\n%s\nwant:\n%s\n", got, want)
-	}
+	//pytest.RegenGoldFiles = true
```
sbinet, line 14: please remove.

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+
+	// GoldFileSuffix is the default expected suffix for a "golden" output file
+	GoldFileSuffix = "_golden.txt"
+)
+
+type TestTask struct {
+	TaskNum  int32                      // Assigned when this task is run
```
sbinet, line 187: is it worth exposing this detail?
it seems only used internally.
```suggestion
	num  int32                      // Assigned when this task is run
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+	TestID   string                     // unique key identifying this task.  If nil, this is autogeneratoed from a PyFile derivative
```
sbinet, line 188:
```suggestion
	ID   string                     // unique key identifying this task.  If empty, this is autogenerated from a PyFile derivative
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+	PyFile   string                     // If set, this file pathname is executed in a newly created ctx
+	PyTask   func(ctx py.Context) error // If set, a new created ctx is created and this blocks until completion
+	GoldFile string                     // Filename containing the "gold standard" stdout+stderr.  If nil, autogenerated from PyFile and TestID
+	Error    error                      // Non-nil if a fatal error is encountered with this task
```
sbinet, line 192:
```suggestion
	Err    error                      // Non-nil if a fatal error is encountered with this task
```

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+		}
+	}
+
+	task.TaskNum = atomic.AddInt32(&taskCounter, 1)
+	if task.TestID == "" {
+		if fileBase == "" {
+			task.TestID = fmt.Sprintf("task-%04d", atomic.AddInt32(&taskCounter, 1))
+		} else {
+			task.TestID = strings.TrimPrefix(fileBase, "./")
+			//nameID = strings.ReplaceAll(fileBase, "/", "_")
```
sbinet, line 213: please remove

examples/embedding/main_test.go
```diff
@@ -5,58 +5,12 @@
...
+	pytest.RunScript(t, "./mylib-demo.py")
```
sbinet, line 15: replacing `TestEmbeddedExample()` with, effectively, `pytest.RunScript(...)` seems to completely discard compiling a customized `gpython` interpreter with the `mylib` module, doesn't it?

if my understanding is correct, then we probably have a problem with that test succeeding :}

pytest/pytest.go
```diff
@@ -132,58 +134,155 @@ func RunBenchmarks(b *testing.B, testDir string) {
...
+	taskCounter int32
+
+	// RegenGoldFiles will cause RunTestTasks() and RunScript() to overwrite their output as the "golden" file (rather than compare against it)
+	RegenGoldFiles bool
+
+	// GoldFileSuffix is the default expected suffix for a "golden" output file
+	GoldFileSuffix = "_golden.txt"
+)
+
+type TestTask struct {
```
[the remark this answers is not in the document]
drew-512: haha yes, good point

examples/embedding/main_test.go
drew-512: Hi friend!  Hope you are well in this crazy world.  :\

Hmm, trying to understand you question here.

My intent with this PR was to more expose and use the "_golden.txt" code you added a couple months ago in order to:
   - drop the (now redundant) golden code in the mylib example
   - lay the groundwork (for a PR to follow) that runs the multi-context example also in the form of a test (as we had previously discussed)

Perhaps to your point, indeed, running `mylib` in a test (vs its `main.go`) is essentially the same.  So if you are saying what is the point of having `main.go`, I would say it's to demonstrate and provide an example that newcomers can relate to, drill into, and not get intimidated.  This is based on my assumption that others are like me and tend to look for things like `main.go` as a starting point (and not tests).   My logic centers around making it as easy and clear as possible for newcomers to explore a completely stand-alone embedded example.

Or perhaps I am misunderstanding your question?

Separately...

Yes, stuff like `pytest.RegenGoldFiles = true` is annoying to have around.  It sounds like the ` flag.Bool(..., false, ...)` approach you described is best.

drew-512:
Lmk what works from here.    

I'm thinking this PR is now good (unless I missed something in your above comments), or I can expand this PR and add the testing for the multi-context test which makes uses of `pytest.Task`.

examples/embedding/main_test.go
sbinet: let me restate my question then.

in the new version of `TestEmbeddedExample`, where is the code of `mylib` coming from?
`mylib-demo.py` imports `mylib`.
and `mylib` imports, in turn, `mylib_go`.

in the old version, `mylib_go` was coming from the compilation of `mylib.module.go` together with `gpython`'s `main`.
now, where is it coming from?

drew-512:
Was holding off on the next branch to do the next branch that tests multi-ctx.  Should I keep waiting or start adding that to this branch?

sbinet:
sorry, I am back from holidays.
taking a look at it now.

sbinet:
there were too many rebase conflicts.

I have created https://github.com/go-python/gpython/pull/192.

sbinet:
superseded by https://github.com/go-python/gpython/pull/192.

PR #215, opened 2023-01-12
Adding __delitem__ to dict

kellrott:
Enables the 'del' command to delete items from a dict

### DIFF
--- a/py/dict.go
+++ b/py/dict.go
@@ -191,14 +191,15 @@ func (d StringDict) M__getitem__(key Object) (Object, error) {
 
 func (d StringDict) M__delitem__(key Object) (Object, error) {
 	str, ok := key.(String)
-	if ok {
-		_, ok := d[string(str)]
-		if ok {
-			delete(d, string(str))
-			return None, nil
-		}
+	if !ok {
+		return nil, ExceptionNewf(KeyError, "%v", key)
 	}
-	return nil, ExceptionNewf(KeyError, "%v", key)
+	_, ok = d[string(str)]
+	if !ok {
+		return nil, ExceptionNewf(KeyError, "%v", key)
+	}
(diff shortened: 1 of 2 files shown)


py/dict.go
```diff
@@ -189,6 +189,18 @@ func (d StringDict) M__getitem__(key Object) (Object, error) {
...
+func (d StringDict) M__delitem__(key Object) (Object, error) {
+	str, ok := key.(String)
+	if ok {
+		_, ok := d[string(str)]
+		if ok {
+			delete(d, string(str))
+			return None, nil
+		}
+	}
+	return nil, ExceptionNewf(KeyError, "%v", key)
```
sbinet, line 201:
```suggestion
	str, ok := key.(String)
	if !ok {
		return nil, ExceptionNewf(KeyError, "%v", key)
	}
	_, ok = d[string(str)]
	if !ok {
		return nil, ExceptionNewf(KeyError, "%v", key)
	}
	delete(d, string(str))
	return None, nil
```

yes, it's longer, but only in the vertical dimension (however, it's less indented)

py/tests/dict.py
```diff
@@ -46,6 +46,12 @@
         assert v == 5.5
 assertRaises(TypeError, a.items, 'a')
 
+doc="del"
+a = {'hello': 'world', 'hi': 'there'}
+del a["hello"]
+assert not a.__contains__('hello')
+assert a.__contains__('hi')
+
```
sbinet, line 54: perhaps add a couple of cases that exercize the 2 `ExceptionNewf` branches ?

sbinet (commented):
thanks for the PR.

I have 2 comments.
see below.

sbinet (approved):
LGTM

(thanks again)

PR #216, opened 2023-01-16
Add os.listdir

reyoung:


reyoung:
@sbinet @glaukiol1 

Please take a review, thanks

stdlib/os/os.go
```diff
@@ -61,6 +61,7 @@ func init() {
 		py.MustNewMethod("rmdir", rmdir, 0, rmdir_doc),
 		py.MustNewMethod("system", system, 0, "Run shell commands, prints stdout directly to default"),
 		py.MustNewMethod("unsetenv", unsetenv, 0, "Unset (delete) the environment variable named key."),
+		py.MustNewMethod("listdir", listDir, 0, listDir_doc),
```
sbinet, line 64: please keep the slice ordered by method name.

(yes, `chdir` should probably be moved up closed to `close`.)

stdlib/os/os.go
```diff
@@ -517,3 +518,59 @@ func system(self py.Object, args py.Tuple) (py.Object, error) {
...
+
+The list is in arbitrary order.  It does not include the special
+entries '.' and '..' even if they are present in the directory.
+`
+
+func listDir(self py.Object, args py.Tuple, kwargs py.StringDict) (py.Object, error) {
+	var (
+		path py.Object = py.None
+	)
+	err := py.ParseTupleAndKeywords(args, kwargs, "s:listdir", []string{"path"}, &path)
```
sbinet, line 538: if we accept strings or bytes-like objects, I believe the argument should be `s*` instead of `s`.
and as we also accept `None`, I believe it should actually be `z*`.

https://docs.python.org/3.11/c-api/arg.html#strings-and-buffers

stdlib/os/testdata/test.py
```diff
@@ -181,6 +182,7 @@
         print("INVALID error caught: %s" % e)
     os.remove(fname)
     os.rmdir(dir2)
+    print(os.listdir(top))
```
sbinet, line 185: please add a test where `top` is of type `bytes-like`.

sbinet: also, please add a test where the argument of `listdir` is `None` (after having `chdir`ed to _e.g._ `top`)

sbinet (changes requested):
thanks for the PR.

I have a couple of comments.
see below.

stdlib/os/testdata/test.py
reyoung: Fixed.

stdlib/os/os.go
reyoung: Fixed.

reyoung:
Perhaps `py.ParseTupleAndKeywords` needs more comments. It is hard to understand how to set the `format` parameter. 

:-)

reyoung:
Oh, I see..

`py.ParseTupleAndKeywords` is just like CPython API `ParseTupleAndKeywords`.

Perhaps need a comment to show that.

sbinet (approved):
LGTM.

could you please send another PR against [go-python/license](https://github.com/go-python/license), adding yourself to the `AUTHORS` and/or `CONTRIBUTORS` files ?
thanks.

reyoung:
Here is the PR for contributors https://github.com/go-python/license/pull/20 .

PR #217, opened 2023-01-17
Updating dict intialization method

kellrott:
Allows for dict initialization using:
 - Keyword arguments: `dict(a="1", b="2", c="3")`
 - Lists of tuples: `dict( zip( "a,b,c".split(","), "1,2,3".split(",") ) )`

### DIFF
--- a/py/dict.go
+++ b/py/dict.go
@@ -9,7 +9,9 @@
 
 package py
 
-import "bytes"
+import (
+	"bytes"
+)
 
 const dictDoc = `dict() -> new empty dictionary
 dict(mapping) -> new dictionary initialized from a mapping object's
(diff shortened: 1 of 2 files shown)


py/dict.go
```diff
@@ -97,6 +99,37 @@ func init() {
 // Used for variables etc where the keys can only be strings
 type StringDict map[string]Object
 
+// DictNew
+func DictNew(metatype *Type, args Tuple, kwargs StringDict) (Object, error) {
+	if len(args) > 1 {
+		return nil, ExceptionNewf(TypeError, "dict expects at most one argument")
```
sbinet, line 105: perhaps add a test that exercizes this case ?

py/dict.go
```diff
@@ -97,6 +99,37 @@ func init() {
...
+	}
+	out := NewStringDict()
+	if len(args) == 1 {
+		arg := args[0]
+		seq, err := SequenceList(arg)
+		if err != nil {
+			return nil, err
```
sbinet, line 112: perhaps add a test that exercizes this case ?

sbinet (commented):
thanks for the PR.
I have some comments, see below.

sbinet (approved):
LGTM

thanks again.

PR #219, opened 2023-03-12
parser: fix CRLF(\r\n) file parsing error, SyntaxError: 'invalid syntax'

wetor:
The python file on the Windows platform may be in CRLF format. For example, git may convert the file to CRLF under Windows, resulting in parsing failure

### DIFF
--- a/parser/lexer.go
+++ b/parser/lexer.go
@@ -110,6 +110,9 @@ func (x *yyLex) dequeue() int {
 func (x *yyLex) refill() {
 	var err error
 	x.line, err = x.reader.ReadString('\n')
+	if strings.HasSuffix(x.line, "\r\n") {
+		x.line = x.line[:len(x.line)-2] + "\n"
+	}
 	if yyDebug >= 2 {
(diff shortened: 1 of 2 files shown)


ncw:
The fix looks reasonable to me.

Can you link some docs showing .py files can end CR LF please?

wetor:
here https://docs.python.org/3.5/reference/lexical_analysis.html
of course, we can also use "build tag" to do special processing on the Windows platform

ncw:
The docs you linked are quite clear

> A physical line is a sequence of characters terminated by an end-of-line sequence. In source files, any of the standard platform line termination sequences can be used - the Unix form using ASCII LF (linefeed), the Windows form using the ASCII sequence CR LF (return followed by linefeed), or the old Macintosh form using the ASCII CR (return) character. All of these forms can be used equally, regardless of platform.

So this should be done on any platform, so we don't need build tags.

Also, theoretically, we should be looking at CR terminated lines. However that is a bigger change to the lexer, so let's not do that until someone requests it as I don't think macs have used CR terminated files for a very long time.

I'll merge this now - thank you :-)

PR #221, opened 2023-03-18
py: int() default to decimal

wetor:
https://docs.python.org/3.4/library/functions.html#int

> int(x, base=10)

If base=0, an exception will be thrown when executing `int("07")`

### DIFF
--- a/py/int.go
+++ b/py/int.go
@@ -55,7 +55,7 @@ func (o Int) Type() *Type {
 func IntNew(metatype *Type, args Tuple, kwargs StringDict) (Object, error) {
 	var xObj Object = Int(0)
 	var baseObj Object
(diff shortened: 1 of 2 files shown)


ncw:
Does this change break this (which is correct)?

```
$ gpython
[snip]
>>> 
>>> 07
Compile error: 
  File "<string>", line 1, offset 0
    07


SyntaxError: 'illegal decimal with leading zero'
>>> 
```

wetor:
I tried it and it didn't affect here, because this PR fixes `ValueError: "invalid literal for int() with base 10: '07'" at "sigils"` error.

ncw (approved):
Looks good to me - thank you :-)

PR #222, opened 2023-06-03
all: fix iterable object, implement `filter`, `map`, `oct` and optimise `hex`

wetor:
## [all: support filter builtin feature and fix iterable object](https://github.com/go-python/gpython/commit/80d61cff6b58f041d7d88486133beb5cd9cfeebf)

Python-3.4.9/Doc/c-api/iterator.rst
> Python provides two general-purpose iterator objects.  The first, a sequence
iterator, works with an arbitrary sequence supporting the :meth:`__getitem__`
method.  The second works with a callable object and a sentinel value, calling
the callable for each item in the sequence, and ending the iteration when the
sentinel value is returned.

`Python-3.4.9/Lib/test/test_builtin.py`
```python
class Squares:

    def __init__(self, max):
        self.max = max
        self.sofar = []

    def __len__(self): return len(self.sofar)

    def __getitem__(self, i):
        if not 0 <= i < self.max: raise IndexError
        n = len(self.sofar)
        while n <= i:
            self.sofar.append(n*n)
            n += 1
        return self.sofar[i]
```
In CPython, `Squares (5)` is an iterable object, and to support this type of iterable object, I made modifications to `Iterator. go`

---

Iterator. go
```go
type Iterator struct {
	Pos int
	Seq Object
}
// ...
func (it *Iterator) M__next__() (res Object, err error) {
	if tuple, ok := it.Seq.(Tuple); ok {
		if it.Pos >= len(tuple) {
			return nil, StopIteration
		}
		res = tuple[it.Pos]
		it.Pos++
		return res, nil
	}
	index := Int(it.Pos)
	if I, ok := it.Seq.(I__getitem__); ok {
		res, err = I.M__getitem__(index)
	} else if res, ok, err = TypeCall1(it.Seq, "__getitem__", index); !ok {
		return nil, ExceptionNewf(TypeError, "'%s' object is not iterable", it.Type().Name)
	}
	if err != nil {
		if IsException(IndexError, err) {
			return nil, StopIteration
		}
		return nil, err
	}
	it.Pos++
	return res, nil
}
```
Corresponding CPython code  
`Python-3.4.9/Objects/iterobject.c`
```c
static PyObject *
iter_iternext(PyObject *iterator)
{
    seqiterobject *it;
    PyObject *seq;
    PyObject *result;

    assert(PySeqIter_Check(iterator));
    it = (seqiterobject *)iterator;
    seq = it->it_seq;
    if (seq == NULL)
        return NULL;
    if (it->it_index == PY_SSIZE_T_MAX) {
        PyErr_SetString(PyExc_OverflowError,
                        "iter index too large");
        return NULL;
    }

    result = PySequence_GetItem(seq, it->it_index);
    if (result != NULL) {
        it->it_index++;
        return result;
    }
    if (PyErr_ExceptionMatches(PyExc_IndexError) ||
        PyErr_ExceptionMatches(PyExc_StopIteration))
    {
        PyErr_Clear();
        Py_DECREF(seq);
        it->it_seq = NULL;
    }
    return NULL;
}
```
---
and `Iter()`
```go
func Iter(self Object) (res Object, err error) {
	if I, ok := self.(I__iter__); ok {
		return I.M__iter__()
	} else if res, ok, err = TypeCall0(self, "__iter__"); ok {
		return res, err
	}
	if ObjectIsSequence(self) {
		return NewIterator(self), nil
	}
	return nil, ExceptionNewf(TypeError, "'%s' object is not iterable", self.Type().Name)
}
```
Corresponding CPython code  
`Python-3.4.9/Objects/abstract.c`
```c
PyObject *
PyObject_GetIter(PyObject *o)
{
    PyTypeObject *t = o->ob_type;
    getiterfunc f = NULL;
    f = t->tp_iter;
    if (f == NULL) {
        if (PySequence_Check(o))
            return PySeqIter_New(o);
        return type_error("'%.200s' object is not iterable", o);
    }
    else {
        PyObject *res = (*f)(o);
        if (res != NULL && !PyIter_Check(res)) {
            PyErr_Format(PyExc_TypeError,
                         "iter() returned non-iterator "
                         "of type '%.100s'",
                         res->ob_type->tp_name);
            Py_DECREF(res);
            res = NULL;
        }
        return res;
    }
}
```

Finally, `filter` and `map` methods were fully implemented

`py/tests/filter.py` is a copy of `Python-3.4.9/Lib/test/test_builtin.py:BuiltinTest.test_filter()`
and
`py/tests/map.py` is a copy of `Python-3.4.9/Lib/test/test_builtin.py:BuiltinTest.test_map()`

## [builtin: Implement oct and optimise hex](https://github.com/go-python/gpython/commit/aa3e6b18adf4981ebb08c7dacb380070343bea79)

Through benchmark testing, the speed of `hex` has indeed been improved  

My English is not very good, so forgive me for not providing sufficient explanations. Welcome to ask questions here

### DIFF
--- a/py/object.go
+++ b/py/object.go
@@ -23,24 +23,53 @@ func ObjectRepr(o Object) Object {
 }
 
 // Return whether the object is True or not
-func ObjectIsTrue(o Object) bool {
-	if o == True {
+func ObjectIsTrue(o Object) (cmp bool, err error) {
+	switch o {
+	case True:
+		return true, nil
+	case False:
+		return false, nil
+	case None:
+		return false, nil
+	}
+
+	var res Object
+	switch t := o.(type) {
+	case I__bool__:
+		res, err = t.M__bool__()
+	case I__len__:
+		res, err = t.M__len__()
+	case *Type:
+		var ok bool
+		if res, ok, err = TypeCall0(o, "__bool__"); ok {
+			break
+		}
+		if res, ok, err = TypeCall0(o, "__len__"); ok {
+			break
+		}
+		_ = ok // pass static-check
+	}
+	if err != nil {
+		return false, err
+	}
+
+	switch t := res.(type) {
+	case Bool:
+		return t == True, nil
+	case Int:
+		return t > 0, nil
+	}
+	return true, nil
+}
+
+// Return whether the object is a sequence
+func ObjectIsSequence(o Object) bool {
+	switch t := o.(type) {
+	case I__getitem__:
 		return true
-	}
-	if o == False {
-		return false
-	}
-
-	if o == None {
-		return false
-	}
-
-	if I, ok := o.(I__bool__); ok {
-		cmp, err := I.M__bool__()
-		if err == nil && cmp == True {
+	case *Type:
+		if t.GetAttrOrNil("__getitem__") != nil {
 			return true
-		} else if err == nil && cmp == False {
-			return false
 		}
 	}
 	return false
--- a/py/arithmetic.go
+++ b/py/arithmetic.go
@@ -147,24 +147,6 @@ func MakeFloat(a Object) (Object, error) {
 	return nil, ExceptionNewf(TypeError, "unsupported operand type(s) for float: '%s'", a.Type().Name)
 }
 
-// Iter the python Object returning an Object
-//
-// Will raise TypeError if Iter can't be run on this object
-func Iter(a Object) (Object, error) {
-
-	if A, ok := a.(I__iter__); ok {
-		res, err := A.M__iter__()
-		if err != nil {
-			return nil, err
-		}
-		if res != NotImplemented {
-			return res, nil
-		}
-	}
-
-	return nil, ExceptionNewf(TypeError, "unsupported operand type(s) for iter: '%s'", a.Type().Name)
-}
-
 // Add two python objects together returning an Object
 //
 // Will raise TypeError if can't be add can't be run on these objects
--- a/py/dict.go
+++ b/py/dict.go
@@ -36,7 +36,7 @@ func init() {
 			return nil, err
 		}
 		sMap := self.(StringDict)
-		o := make([]Object, 0, len(sMap))
+		o := make(Tuple, 0, len(sMap))
 		for k, v := range sMap {
 			o = append(o, Tuple{String(k), v})
 		}
@@ -49,7 +49,7 @@ func init() {
 			return nil, err
 		}
 		sMap := self.(StringDict)
-		o := make([]Object, 0, len(sMap))
+		o := make(Tuple, 0, len(sMap))
 		for k := range sMap {
 			o = append(o, String(k))
 		}
@@ -62,7 +62,7 @@ func init() {
 			return nil, err
 		}
 		sMap := self.(StringDict)
-		o := make([]Object, 0, len(sMap))
+		o := make(Tuple, 0, len(sMap))
 		for _, v := range sMap {
 			o = append(o, v)
 		}
@@ -204,7 +204,7 @@ func (a StringDict) M__repr__() (Object, error) {
 
 // Returns a list of keys from the dict
 func (d StringDict) M__iter__() (Object, error) {
-	o := make([]Object, 0, len(d))
+	o := make(Tuple, 0, len(d))
 	for k := range d {
 		o = append(o, String(k))
 	}
--- a/py/exception.go
+++ b/py/exception.go
@@ -336,6 +336,8 @@ func ExceptionGivenMatches(err, exc Object) bool {
 func IsException(exception *Type, r interface{}) bool {
 	var t *Type
 	switch ex := r.(type) {
+	case ExceptionInfo:
+		t = ex.Type
 	case *Exception:
 		t = ex.Type()
 	case *Type:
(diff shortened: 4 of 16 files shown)


py/object.go
```diff
@@ -23,24 +23,50 @@ func ObjectRepr(o Object) Object {
...
 		cmp, err := I.M__bool__()
-		if err == nil && cmp == True {
+		if err != nil {
+			return false, err
+		}
+		if cmp == True {
+			return true, nil
+		} else if cmp == False {
+			return false, nil
+		}
```
sbinet, line 47:
```suggestion
		switch cmp {
		case True:
			return true, nil
		default:
			return false, nil
		}
```

py/object.go
```diff
@@ -23,24 +23,50 @@ func ObjectRepr(o Object) Object {
...
+	}
+	if I, ok := o.(I__len__); ok {
+		l, err := I.M__len__()
+		if err != nil {
+			return false, err
+		}
+		if l.(Int) > 0 {
+			return true, nil
+		} else {
+			return false, nil
```
sbinet, line 57:
```suggestion
		switch {
		case l.(Int) > 0:
			return true, nil
		default:
			return false, nil
```

py/object.go
```diff
@@ -23,24 +23,50 @@ func ObjectRepr(o Object) Object {
...
+		}
+	}
+	return false, nil
```
sbinet, line 60: we should probably try to add a test for this case.

py/object.go
sbinet: and add a test for this `return false, nil` case

py/object.go
```diff
@@ -23,24 +23,50 @@ func ObjectRepr(o Object) Object {
...
+}
+
+// Return whether the object is True or not
+func ObjectIsSequence(o Object) bool {
+	if _, ok := o.(I__getitem__); ok {
+		return true
+	} else if t, ok := o.(*Type); ok {
+		if t.GetAttrOrNil("__getitem__") != nil {
```
sbinet, line 68:
```suggestion
	switch t := o.(type) {
	case I__getitem__:
		return true
	case *Type:
		if t.GetAttrOrNil("__getitem__") != nil {
```

sbinet (commented):
thanks for the PR.

I have mostly cosmetics comments, see below.

py/object.go
wetor: `ObjectIsTrue` has been written in a different way, which may look more beautiful, but it may also not fit some checklists,  please review again. At the same time, testing coverage for this method has been added in `filter.py`

sbinet: your rewrite seems to be bug-compliant with the original write up :)

according to the python ref:
"""
[T]he following values are interpreted as false: False, None, numeric zero of all types, and empty strings and containers (including strings, tuples, lists, dictionaries, sets and frozensets). All other values are interpreted as true. 
"""
https://docs.python.org/3/reference/expressions.html#booleans

so with both your rewrite and the previous version, the following script:
```python
class F(object):
 pass

f = F()
if f:
 print("true")

print(all([f]))
```
leads to:
```
true
False
```
whereas with python3 (3.11) this leads to:
```
true
True
```

you may or may not want to fix this as part of this CL.

sbinet: the bug may actually comes from a wrong use of `ObjectIsTrue`: it may be because this function is not meant to be mirroring the python's truthness semantics.

@ncw ? do you confirm/infirm ?

wetor: This is a simple bug
```c
int
PyObject_IsTrue(PyObject *v)
{
    Py_ssize_t res;
    if (v == Py_True)
        return 1;
    if (v == Py_False)
        return 0;
    if (v == Py_None)
        return 0;
    else if (v->ob_type->tp_as_number != NULL &&
             v->ob_type->tp_as_number->nb_bool != NULL)
        res = (*v->ob_type->tp_as_number->nb_bool)(v);
    else if (v->ob_type->tp_as_mapping != NULL &&
             v->ob_type->tp_as_mapping->mp_length != NULL)
        res = (*v->ob_type->tp_as_mapping->mp_length)(v);
    else if (v->ob_type->tp_as_sequence != NULL &&
             v->ob_type->tp_as_sequence->sq_length != NULL)
        res = (*v->ob_type->tp_as_sequence->sq_length)(v);
    else
        return 1;
    /* if it is negative, it should be either -1 or -2 */
    return (res > 0) ? 1 : Py_SAFE_DOWNCAST(res, Py_ssize_t, int);
}
```
The default return value for CPython is 1, and I only need to modify the default return value to fix it

wetor: The correct logic should be:

If it is `None` or `False`, return `False`;

If implemented`__bool__` or `__Len__` then return directly, where `__Len__ > 0` is True;

If an error occurs, return `False`;

Otherwise, return `True`;

wetor:
Of course, I will split commit later

wetor:
I found that the basic types (int, string, etc.) did not run `Ready()` correctly, resulting in`__doc__` The attribute was not registered correctly and was fixed by the way

sbinet (approved):
LGTM, modulo a very minor nit.

thanks again.

sbinet (approved):
LGTM.

thanks again.

PR #226, opened 2023-11-22
Fixed WASM compatibility by avoiding panic due to absence of `os.Executable()`

vasilev:
`os.Executable()` is not implemented in "js" and "wasip1" targets in current Golang versions, so WASM builds encounter `Executable not implemented for js` errors in runtime.

This fix is borrowed from Brad's https://github.com/tailscale/tailscale/pull/8325 .

### DIFF
--- a/stdlib/sys/sys.go
+++ b/stdlib/sys/sys.go
@@ -19,6 +19,7 @@ package sys
 
 import (
 	"os"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


ncw (approved):
This is a nice fix - thank you - I'll merge it now.

vasilev:
Thank you too!

PR #241, opened 2025-06-28
py: fix import on Windows

aisk:
Currently gpython can't handle import on Windows because it's using `path` to handle file paths, which only support OSes which use `/` as path separator:

```
PS C:\Users\xxxxx\Source\gpython> go run . py\tests\string.py
  File "C:\\Users\\xxxxx\\Source\\gpython/py\\tests\\string.py", line 5, in <module>
FileNotFoundError: 'Failed to resolve "libtest"'
exit status 1
```

Using `path/filepath` can fix it.

### DIFF
--- a/py/import.go
+++ b/py/import.go
@@ -7,7 +7,7 @@
 package py
 
 import (
-	"path"
+	"path/filepath"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


ncw (approved):
Looks great - thanks :-)

PR #245, opened 2025-07-04
pytest: ensure expected string's line sep is \n

aisk:
Some tests failed on my Windows machine, and I found that some excepted output are using "\r\n" as line separator. The tests passed on Github Actions's Windows environment, so I guess git have some configs that will convert the line separator to the default one in current OS ("\r\n" on Windows).

So I think we still need to fix it, because someone else may have this git config too.


ncw (approved):
Looks like a good idea. Thank you.

PR #256, opened 2026-03-19
Add runner context interrupting

TOMMy-Net:
Adding context interrupting based on atomics, for breaking code runner.

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,7 +14,6 @@ import (
 	"os/signal"
 	"runtime"
 	"runtime/pprof"
-	"syscall"
 
 	"github.com/go-python/gpython/py"
 	"github.com/go-python/gpython/repl"
@@ -51,7 +50,7 @@ func xmain(args []string) {
 	defer ctx.Close()
 
 	sigCh := make(chan os.Signal, 1)
-	signal.Notify(sigCh, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(sigCh, os.Interrupt)
 	go func() {
 		for range sigCh {
 			ctx.SetInterrupt()
--- a/main.go
+++ b/main.go
@@ -11,8 +11,10 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
 	"runtime"
 	"runtime/pprof"
+	"syscall"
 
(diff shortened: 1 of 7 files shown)


stdlib/stdlib.go
```diff
@@ -44,6 +45,7 @@ type context struct {
 	closed    bool
 	running   sync.WaitGroup
 	done      chan struct{}
+	interrupt int32 // atomic; non-zero means KeyboardInterrupt pending
```
ncw: Using atomic.Int32 here saves mistakes

main.go
```diff
@@ -48,6 +50,14 @@ func xmain(args []string) {
 	ctx := py.NewContext(opts)
 	defer ctx.Close()
 
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
ncw: This won't work on windows I think.

We probably only need os.Interrupt anyway, INT and TERM are not generated by the keyboard.

Perhaps see what python3 does with those signals?

ncw (changes requested):
This looks like a nice change - thank you

See inline for comments

stdlib/stdlib.go
TOMMy-Net: Go 1.8, not support this type of atomics @ncw

main.go
TOMMy-Net: "`INT and TERM are not generated by the keyboard.`"   You sure? I thought that INT was called by pressing CTRL+C.

ncw: You are right of course. However os.Interrrupt is the same as sigint on Unix machines, so it should cover everything we need

stdlib/stdlib.go
ncw: Good point! I forgot we are still on 1.8. Maybe we should upgrade the minimum requirements. What do you think?

TOMMy-Net: > Good point! I forgot we are still on 1.8. Maybe we should upgrade the minimum requirements. What do you think?

I don't think anything will break). I have tested. @ncw

TOMMy-Net:
What about this changes??
@ncw

